PR #924, opened 2015-02-21
Fix scripts

yavorsky:


es128:
No, why? We want the code compiled prior to publishing to npm, no need to have every user do it locally.

yavorsky:
@es128 Ok, but in .gitignore I found `lib/`. So, after `npm i -g brunch` there isn't lib/ in `/usr/local/lib/node_modules/brunch`.

es128:
It isn't in npmignore though. Why is gitignore relevant?

yavorsky:
@es128  Yes, of course, but there is problems when we want to use edge version from github. ex. `sudo npm install -g https://github.com/brunch/brunch/tarball/master`.

es128:
Ah that's right. I had forgotten why we had it this way. Still, perhaps the `postinstall` could be set to run `preinstall` if `lib` is absent, otherwise noop.

paulmillr:
we had the good setup.js script — perphaps we need to bring it back.

PR #925, opened 2015-02-21
Add setup

yavorsky:
Return setup.js, fix package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,8 +27,10 @@
     "brunch": "./bin/brunch"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


es128:
Is there not a way to check for `lib` with a simple shell command? Even if not, I think we were getting closer to an improvement before and we should minimize what goes into the setup.js to only the `postinstall` part that can't easily be accomplished otherwise.

paulmillr:
i don't think so; in a cross-platform way

PR #926, opened 2015-02-21
Add postinstall script, fix chokidar version

yavorsky:


es128:
I like it, but have you tried it on windows?

package.json
```diff
@@ -28,6 +28,7 @@
   },
   "scripts": {
     "prepublish": "coffee -o lib/ src/",
+    "postinstall": "if [ ! -d lib ]; then coffee -o lib/ src/; fi",
```
es128: How about `npm run prepublish` here instead of the `coffee` command?

yavorsky: @es128 :+1:  But need to modify script for windows...

es128: Perhaps something like `node -e 'if (fs.existsSync(...`

yavorsky:
fs.exists was deprecated. I'll fix with alternative in a minute

es128:
I know it's deprecated, but for the sake of brevity wouldn't mind if it was used here in this way.

es128:
Are you sure the `require` statements are needed with `node -e`? It's not necessary in the REPL

yavorsky:
@es128 necessary in eval mode.

es128:
Ok. Too bad.

es128:
```json
"postinstall": "node -e 'require(\"fs\").readdir(\"lib\",function(e){e&&require(\"child_process\").exec(\"npm run prepublish\")})'"
```

es128:
This appears to be ready. Any other feedback?

yavorsky:
I thinks it's the shortest possible cross-platform script

paulmillr:
have you tested it on windows?

es128:
I did just try it on windows, and it looks like the only issue is that windows requires an inversion of the double and single quotes (more specifically, the cli string argument must be wrapped in double-quotes, it doesn't understand the single quotes there)

`"postinstall": "node -e \"require('fs').readdir('lib',function(e){e&&require('child_process').exec('npm run prepublish')})\""`

@yavorsky Please set it like this in your branch and I'll try again to install from the remote git source just to be sure.

paulmillr:
@es128 is it cool?

es128:
Seems ok

image

PR #930, opened 2015-02-28
Implement node_modules support.

yavorsky:
This PR is needed to discus some aspects. For now, it is working good with files from npm like 'backbone' or 'react', also, it ignores brunch plugins, but there is some problems when we have packages which require node's native modules like `fs` or `path` (ex. express) . I solved this by adding /^express/ to `conventions.ignored`. My proposition is to create separate branch for this implementation (maybe for collective polishing).

### DIFF
--- a/src/fs_utils/pipeline.coffee
+++ b/src/fs_utils/pipeline.coffee
@@ -74,12 +74,12 @@ isNpm = (path) ->
 
 pipeline = (path, linters, compilers, callback) ->
   if isNpm path
-    deppack path, {basedir: '.', rollback: true}, (error, source) ->
+    deppack path, {basedir: '.', rollback: true, ignoreRequireDefinition: true}, (error, source) ->
       compile source, path, compilers, callback
   else
     fcache.readFile path, (error, source) =>
-      debug "Linting '#{path}'"
       lint source, path, linters, (error) ->
+        debug "Linting '#{path}'"
         if error?.toString().match /^warn\:\s/i
           logger.warn "Linting of #{path}: #{error}"
         else
--- a/src/fs_utils/pipeline.coffee
+++ b/src/fs_utils/pipeline.coffee
@@ -4,6 +4,9 @@ debug = require('debug')('brunch:pipeline')
 fcache = require 'fcache'
 sysPath = require 'path'
 logger = require 'loggy'
+deps = require 'module-deps'
+pack = require 'browser-pack'
+JSONStream = require 'JSONStream'
 
 throwError = (type, stringOrError) =>
   string = if stringOrError instanceof Error
@@ -68,7 +71,25 @@ compile = (source, path, compilers, callback) ->
   first = (next) -> next null, {source, path, callback}
   waterfall [first].concat(compilers.map mapCompilerChain), callback
 
+isNpm: (path) ->
+  path.indexOf('node_modules') >= 0
+
+readWithDeps: (src, cb) ->
+  srcPath = sysPath.join rootPath, 'node_modules', src
+  srcJson = require sysPath.join srcPath, 'package.json'
+  srcMain = srcJson.main || 'index.js'
+  md = deps()
+  pck = pack()
+  data = ''
+  md.pipe(JSONStream.stringify()).pipe(pck).pipe(process.stdout)
+  md.end({ file: rootPath + '/node_modules/' + src + '/' + srcMain})
+  md.on('end', -> cb(null, data))
+
+
 pipeline = (path, linters, compilers, callback) ->
+  # if isNpm(path)
+
+  # else
   fcache.readFile path, (error, source) =>
     debug "Linting '#{path}'"
     lint source, path, linters, (error) ->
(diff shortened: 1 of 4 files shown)


es128:
Why not browserify for packaging and shimming the native modules?

yavorsky:
I agree to use browserify. It more stable then deppack. 
But: 
1. It uses streams, instead of callbacks (like brunch), so data reading will be smthing like 

```coffeescript
b = browserify().add(fileName)
stream = b.bundle()
allData = ''
stream.on 'data', (data) -> allData += data
stream.on 'end', -> compile allData #...
```
2. Deppack is simpler.
3. Not such stable as browserify, but its temporarily.
4. Bundling time is the same.
5. Deppack is flexible for our needs.
6. The single drawback of deppack is throwing error when loading native modules, but i'm going to fix this.

es128:
Ok sounds good, but as you suggested we should not land this in master until it is more mature.

yavorsky:
@es128 Yes! I think @paulmillr will share his thoughts soon. Also, I need to fix shimming the native modules and we need to solve the problem of loading backend packages. Maybe it's viable to put backend scripts to peerDependencies or devDependencies. :/

es128:
Use `detective` on app code to figure out which node_modules are being `require`d

yavorsky:
@es128 :+1: We [used it](https://github.com/yavorsky/deppack/blob/master/src/deppack.coffee#L46) in deppack, nice stuff, but the native modules can be used in children dependencies, so I think we need to rollback all parents in this cases. Not bad

es128:
Or pull in the browserify shims as needed

es128:
Is deppack reading the app code from the brunch project? Or just using detective to figure out the dep graph? 

If the former, then what's the issue with stuff like express? As long as the user didn't try to `require('express')` in their web app code.

paulmillr:
wait for my review on monday please!

src/fs_utils/pipeline.coffee
```diff
@@ -68,15 +69,23 @@ compile = (source, path, compilers, callback) ->
...
-    lint source, path, linters, (error) ->
-      if error?.toString().match /^warn\:\s/i
-        logger.warn "Linting of #{path}: #{error}"
-      else
-        return callback throwError 'Linting', error if error?
+isNpm = (path) ->
+  path.indexOf('node_modules') >= 0
 
+pipeline = (path, linters, compilers, callback) ->
+  if isNpm(path)
```
paulmillr: drop the `()`

src/fs_utils/pipeline.coffee
```diff
@@ -68,15 +69,23 @@ compile = (source, path, compilers, callback) ->
...
+    deppack path, {basedir: '.'}, (err, source) ->
+      console.log path
       compile source, path, compilers, callback
+  else
+    fcache.readFile path, (error, source) =>
+      debug "Linting '#{path}'"
```
paulmillr: We should move the debug statement inside `lint` fn

paulmillr:
Landed. Works only when Brunch was ran through `brunch.coffee` dev binary though. So, we can safely make a new release.

damusnet:
`./node_modules/brunch/bin/brunch.coffee build` works fine, but `./node_modules/brunch/bin/brunch.coffee watch` fails as follows:
```
/myapp/node_modules/brunch/node_modules/source-map/lib/source-map/source-node.js:100
            var code = nextLine.substr(0, mapping.generatedColumn -
                                ^
TypeError: Cannot call method 'substr' of undefined
  at Function.<anonymous> (/myapp/node_modules/brunch/node_modules/source-map/lib/source-map/source-node.js:100:33)
  at Array.forEach (native)
  at SourceMapConsumer_eachMapping [as eachMapping] (/myapp/node_modules/brunch/node_modules/source-map/lib/source-map/source-map-consumer.js:167:10)
  at Function.SourceNode_fromStringWithSourceMap [as fromStringWithSourceMap] (/myapp/node_modules/brunch/node_modules/source-map/lib/source-map/source-node.js:84:26)
  at updateMap (/myapp/node_modules/brunch/src/fs_utils/source_file.coffee:29:16)
  at updateCache (/myapp/node_modules/brunch/src/fs_utils/source_file.coffee:58:16)
  at /myapp/node_modules/brunch/src/fs_utils/source_file.coffee:69:7
  at /myapp/node_modules/brunch/src/fs_utils/pipeline.coffee:64:7
  at getDependencies (/myapp/node_modules/brunch/src/fs_utils/pipeline.coffee:36:5)
  at /myapp/node_modules/brunch/src/fs_utils/pipeline.coffee:62:5
  at JavaScriptCompiler.compile (/myapp/node_modules/javascript-brunch/lib/index.js:23:10)
  at /myapp/node_modules/brunch/src/fs_utils/pipeline.coffee:66:20
  at fn (/myapp/node_modules/brunch/node_modules/async-waterfall/index.js:19:24)
  at Object._onImmediate (/myapp/node_modules/brunch/node_modules/async-waterfall/index.js:58:22)
  at processImmediate [as _immediateCallback] (timers.js:330:15)
```

Issue #938, opened 2015-03-22
Build problem 'EISDIR' in Windows


kazelsama:
sorry for the little text, I do not speak English

###### OS
<pre>Windows 7 x64</pre>
###### node version
<pre>v0.12.0</pre>
###### folders
prices
╠ .sass-cache
╠ app
║╠ css
║╠ img
║╠ js
║╠ sass
║╠ views
║╚ <code>index.html</code>
╠ bower_components
╠ build
╠ node_modules
╠ .gitignore
╠ bower.json
╠ brunch-config.js
╠ config.rb
╚ package.json
###### brunch-config.js
<pre>
exports.config={
	"paths":{
		"watched":["app"],
		"public":"build",
		"compass":"config.rb"
	},
	"files":{
		"stylesheets":{
			"joinTo":"css/app.css",
			"order":{
				"after":[/css[\\/]/]
			}
		},
		"javascripts":{
			"joinTo":"js/app.js",
			"order":{
				"after":[/js[\\/]/]
			}
		}
	},
	"conventions":{
		"assets":[/img/,/index.html/]
	},
	"modules":{
		"wrapper":false,
		"definition":false
	},
	"overrides":{
		"DEV":{
			"optimize":true,
			"sourceMaps":true,
			"plugins":{
				"autoReload":{
					"enabled":true
				}
			}
		}
	}
};
exports.startServer={
	"server":{
		"indexPath":"app/"
	}
};
</pre>
###### bower.json
<pre>
{
  "name": "prices",
  "version": "0.0.0",
  "authors": [
    "kazelsama"
  ],
  "description": "description",
  "main": "app/index.html",
  "keywords": [
    "prices"
  ],
  "license": "MIT",
  "private": true,
  "ignore": [
    "**/.*",
    "node_modules",
    "bower_components",
    "test",
    "tests"
  ],
  "dependencies": {
    "jquery": "~2.1.3",
    "angular-material": "~0.8.3",
    "angular-material-icons": "~0.4.0"
  }
}
</pre>
###### error
<pre>
D:\wamp\www\prices>brunch build
22 Mar 13:56:05 - error: Copying of 'app\index.html' failed. Error: EISDIR, open 'D:\wamp\www\prices\build'
</pre>
with verbose
<pre>
brunch:common File copy: Error: EISDIR, open 'D:\wamp\www\prices\build' +43s
</pre>

kazelsama:
I fix it:
in "brunch\lib\fs_utils\common.js"
[...]

PR #939, opened 2015-03-22
closes brunch/brunch#938

kazelsama:
closes gh-938


kazelsama:
in some cases the result is: 'D:\wamp\www\prices\cache\market.html\market.html'

paulmillr:
I have moved the function to https://github.com/paulmillr/quickly-copy-file

Issue #967, opened 2015-04-29
Brunch.build() failing after 1.8.0


matthopson:
I've run into an issue after migrating from 1.7.20 to 1.8.x 

We use a deploy script that runs a build and moves changed files to various environments. As of 1.8, the deploy was failing with:

`chdir.call(process, d)
        ^
Error: Bad argument.
  at process.chdir`

Stepping through it, the issue is appearing in `watch.js` around here: `process.chdir(path)`
Sure enough, `path` is set to `{env: 'staging'}`

I did see this recent change, which seems to be the culprit given my current config.
https://github.com/brunch/brunch/blob/9b9e8c5015856eb8e770f050232e092ee88e822e/src/watch.coffee#L576
It's possible I'm overlooking something obvious and this is not a bug - I should mention if I remove the following, everything works as expected.
`if (path) {
    process.chdir(path);
  }`
  It is reproducible in OSX and from within a Vagrant box. (Node 0.10.33)

Thanks!

paulmillr:
Are you using CLI or the node Api?

es128:
Sounds like you're running something like `brunch build d` instead of `brunch build -d`

es128:
Nevermind, just re-read and think I have a better idea of what's going on. Will work on figuring this out today. Also, appears to be the same bug as #957

PR #974, opened 2015-05-20
Handle optional `path` arg

matthopson:
Hey!

Recently I had an issue with brunch builds failing with versions >= 1.8 (https://github.com/brunch/brunch/issues/967)

The issue was due to the addition of the `path` argument in `watch()`
https://github.com/brunch/brunch/commit/9b9e8c5015856eb8e770f050232e092ee88e822e#diff-5894f29c43782173bc5f20c4d13a4f8d

In my case, it was failing because I had a deployment script that would build on the fly and deploy based on the env (override). This script would call `brunch.build options`, but with the addition of the `path` argument,  path was now being set to my `options` argument. I can get around this by simply adding `brunch.build null, options`. This isn't a difficult fix on my end, but I suspect others might run into the same issue.

This could be cleaned up a bit outside of the watch function, but this PR offers a non-depricating solution to any code currently using the same method as I am for builds.

Feel free to incorporate or improve.

Thanks!

### DIFF
--- a/src/watch.coffee
+++ b/src/watch.coffee
@@ -580,5 +580,11 @@ class BrunchWatcher
     start
 
 module.exports = watch = (persistent, path, options, callback = (->)) ->
-  process.chdir path if path
+  # If path isn't provided (by CL)
+  if path
+    if typeof path is 'string'
+      process.chdir path
+    else
+      callback = options if typeof 'options' is 'function'
+      options = path
   new BrunchWatcher(persistent, options, callback)


es128:
This is the first time I've realized that anyone uses the Brunch API outside of the CLI (didn't recognize that's what you meant from the issue you filed), so I apologize for not having followed semver appropriately here and breaking your build process.

Your PR looks good. Thanks

matthopson:
No worries. I should have been more clear :stuck_out_tongue:  Thanks!!

PR #985, opened 2015-06-16
Documentation Fixes

oscherler:
* `docs/config.md`:
    * Updated the default value for `conventions.vendor`: it’s `/(^bower_components|node_modules|vendor)[\\/]/`, but was said to be `/vendor[\\/]/`, which makes it hard to understand how Bower components are handled;

    * Updated the default value for `conventions.ignored `: added `vendor/ruby-*/`, `vendor/jruby-*/`, and `vendor/bundle/`;

    * Added mention of the `isVendor`argument for the `modules.wrapper` function;

* `docs/commands.md`: Corrected an example still using the old `brunch new` syntax with `--skeleton`;

* `docs/faq.md`: Added a mention of the `--debug` option;

* `docs/README.md`: I believe it’s _Any scripts in `app/` dirs_ that are wrapped by default into modules, and not only _Any scripts in `assets/` dirs_, which looks like a copy/paste mistake.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -78,7 +78,7 @@ via modifying config file.
 Brunch also has conventions. Conventions are filters for files with special meaning. They can be changed via config.
 
 * Static files in `assets/` dirs are copied directly to `public/`.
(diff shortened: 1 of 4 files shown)


paulmillr:
Thanks dude!

Issue #986, opened 2015-06-22
NPM import strategy


warrenseine:
Hi,

I'm very glad NPM support was recently added. Bower was nice for a time, but dealing with dependencies in a single place is _much_ better.

Unfortunately, this does not do what I would have expected. That is, packaging the "distributed" file when it exists. I can see that Brunch is doing its best and reads the `main` attribute of `package.json`. But in most cases, this is not what I want:
* React, a very popular library, comes in a bundled version on NPM, with a ready-to-be-used `dist/react.js` but `main` references `main.js` (a "source" file)
* React Router has `umd/ReactRouter.js` but `main` references `lib` (a directory!)

Can I override this behavior to tell Brunch which file to package for each NPM module?

warrenseine:
Basically, my idea would have been to fix `loadNpm` by looking at the `before` array. If some path matches `node_modules/{dep}/*`, then use this file instead of `package.json`'s `main`. It sounds a bit clumsy so I'm asking for better ideas here.

es128:
Why not point to the specific file you want in the `require` call in your app code? `require('react/dist/react')` or something like that. 

I don't think `before` idea would work out in a general sense. If the above suggestion doesn't work, we may need something more like what happened with bower `overrides`.

gjaldon:
@warrenseine - were you able to use react-router after overriding main with `umd/ReactRouter.js`? I tried it out but requiring reactRouter and using it gives me the error `Uncaught TypeError: _reactRouter2.default.run is not a function`. 

React works fine though when overring main with `dist/react.js`.

warrenseine:
Yes, I figured out that `umd` was not intented to be used as `dist`. Actually, there's an issue [here](https://github.com/rackt/react-router/issues/839) to get a properly `dist` in the repository. You can precompile it yourself anyway.

gjaldon:
@warrenseine I see. What do you mean about precompiling it? Do you precompile it with browserify now? Do you use just NPM? I'm curious as to how you went about this.

warrenseine:
@gjaldon I made a private repo with `dist` built and embedded and pointed my `package.json` to it.

paulmillr:
Let's move to gh-1023.

PR #987, opened 2015-06-23
Add "overrides" support for NPM.

warrenseine:
Similar to the unofficial `overrides` keyword in `bower.json`, this patch adds support for an unofficial `overrides` keyword in `package.json`, in order to fix issue #986.

A supported `package.json` would be:

    {
      ...

      "dependencies": {
        "react": "*"
      },
      "overrides": {
        "react": {
          "main": "dist/react.js"
        }
      }
    }

### DIFF
--- a/src/helpers.coffee
+++ b/src/helpers.coffee
@@ -400,6 +400,7 @@ loadNpm = (config, cb) ->
     .map (dep) ->
       depPath = sysPath.join rootPath, 'node_modules', dep
       depJson = require sysPath.join depPath, 'package.json'
+      depJson = deepExtend(depJson, json.overrides[dep]) if json.overrides[dep]
       depMain = depJson.main or 'index.js'
       file = sysPath.join 'node_modules', dep, depMain
       name: dep


warrenseine:
See related issue #986 and a [blog post](https://roots.io/using-bootstrap-with-bower-how-to-override-bower-packages/) explaining the `overrides` keyword in Bower.

warrenseine:
Note that you can't override package dependencies with this PR (use shrinkwrap for that). It's only a shallow addition that is helpful when packaging your `vendor.js`.

es128:
That's an interesting blog reference regarding `overrides` for bower. I'm pretty sure that's something that got adopted in Brunch first, didn't realize the concept had spread so much.

You didn't really respond regarding the `require('react/dist/react')` idea. Am interested in why that isn't adequate - seems more straightforward than messing with the package.json. In either case, the automagic has failed and you need to specify the precise file(s) you want included in your build. Why split the reference up to two places, i.e. `require('react')` in your app and then `overrides` to specify what you really meant by that.

warrenseine:
I don't like `require('react/dist/react')` because it's mixing semantics (requiring a module) and implementation details (a specific built file).

Second, there are many places where I use React and I don't want to remember which file I need to require. Having this in a central place is safe and clean.

Worst, it might change, or I may want to use `react/dist/react.min.js` in production.

es128:
Thanks for the response. That does sound reasonable.

:+1: on the concept. PR just needs docs.

paulmillr:
👍

warrenseine:
If that's good, can't it be merged for 1.8.4?

es128:
> PR just needs docs.

warrenseine:
There's a separate issue about it. Moreover, the overall npm support needs docs (there is none AFAIK), this is not specific to my PR.

es128:
@warrenseine you're right, but I hesitate to publish until we get docs in place. I'm ok with merging into master for now.

This should probably be a 1.9.0 release (after docs are added). Yes, we've bumped minor when adding new features/options many times in the past, but we should try to correct that. @paulmillr wdyt?

paulmillr:
Makes sense.

ryansobol:
I really need this feature for an educational course I'm developing on React. If I submit a PR with docs on working with NPM dependencies, will brunch v1.9.0 be ready to ship?

es128:
Yes, that would be great, although these changes did already get published in the latest release.

PR #1000, opened 2015-07-29
Support config.modules.appendRequire

### LINKED ISSUE #999: Make brunch play nicely with CSP
I am working on adding Content Security Policy (CSP) rules to Phoenix applications and I have been struggling to make it play nicely with Brunch. The issue is that Brunch requires a script tag to require your main application module::

    <script>require("app")</script>

However CSP recommends to have no inline scripts in HTML pages. So the `require("app")` needs to be automatically embedded in the compiled script. Do you have any ideas in mind to solve this issue? I want to avoid bringing things like browserify though because it is a large dependency for a simple problem.

For example, one solution I had in mind is to support "autorequires", so one could write:

    joinTo:
      'js/app.js':
        include: /^(web\/static\/js)/
        autorequire: ["app"]
      'js/vendor.js':
        include: /^(web\/static\/vendor)/

Is there something like that in Brunch? If not, is a PR desired?

josevalim:
This allow us to customize which files to automatically
require at the end of each joined file.

Closes #999

### DIFF
--- a/src/fs_utils/generate.coffee
+++ b/src/fs_utils/generate.coffee
@@ -61,7 +61,7 @@ sort = (files, config, joinToValue) ->
     indexes[path]
 
 # New.
-concat = (files, path, type, definition, aliases) ->
+concat = (files, path, type, definition, aliases, appendRequire) ->
   # nodes = files.map toNode
   root = new SourceNode()
   debug "Concatenating #{files.map((_) -> _.path).join(', ')} to #{path}"
@@ -75,6 +75,9 @@ concat = (files, path, type, definition, aliases) ->
   aliases?.forEach (alias) ->
     key = Object.keys(alias)[0]
     root.add "require.alias('#{key}', '#{alias[key]}');"
+
+  appendRequire?.forEach (require) ->
+    root.add "require('#{require}');"
 
   root.toStringWithSourceMap file: path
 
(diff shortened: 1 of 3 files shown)


josevalim:
Folks, let me know if this is fine as is or if you want it to be renamed to `autoRequire`.

es128:
Solution LGTM

I don't have a terribly strong opinion about the name.

src/fs_utils/generate.coffee
```diff
@@ -124,7 +127,7 @@ generate = (path, sourceFiles, config, optimizers, callback) ->
   joinToValue = config.files["#{type}s"].joinTo[joinKey]
   sorted = sort sourceFiles, config, joinToValue
 
-  {code, map} = concat sorted, path, type, config._normalized.modules.definition, config._normalized.packageInfo.component.aliases
+  {code, map} = concat sorted, path, type, config._normalized.modules.definition, config._normalized.packageInfo.component.aliases, config._normalized.modules.appendRequire[joinKey]
```
es128: this line needs to be broken up, but maybe not as part of this PR

paulmillr:
Let's do `autoRequire`. We're currently prepending require definition to the file already. So a user may think it's something related, when it's not.

PR #1001, opened 2015-07-29
Support config.modules.autoRequire

### LINKED ISSUE #999: Make brunch play nicely with CSP

josevalim:
This allow us to customize which files to automatically
require at the end of each joined file.

Closes #999. Alternative to #1000.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -122,6 +122,15 @@ require.define({#{path}: function(exports, require, module) {
     """
 ```
 
+`modules.autoRequire`: `Object` specifies requires to be automatically added at the end of joined file. The example below will require both 'app' and 'foo':
+
+```coffeescript
+# Default behaviour.
+modules:
+  autoRequire:
+    'javascripts/app.js': ['app', 'foo']
+```
+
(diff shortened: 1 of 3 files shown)


paulmillr:
Thanks dude. 

@es128 — could you please push a new release? I can do this too, but tomorrow.

PR #1002, opened 2015-07-31
Check whether json.overrides is defined

linkvt:
In #987 the overrides support was added. After updating I noticed, that my build failed because `overrides` wasn't defined in my `package.json`. I think this shouldn't be required.

As I'm not using CoffeeScript regularly, I hope this is the correct way to do it.

### DIFF
--- a/src/helpers.coffee
+++ b/src/helpers.coffee
@@ -402,7 +402,7 @@ loadNpm = (config, cb) ->
     .map (dep) ->
       depPath = sysPath.join rootPath, 'node_modules', dep
       depJson = require sysPath.join depPath, 'package.json'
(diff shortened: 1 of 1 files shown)


es128:
Seems alright. Did you try using the patch for your own build and confirm it resolved the issue?

linkvt:
Yes I followed the steps of your contributing guidelines. 
You can reproduce this by creating a new project from the simple skeleton, adding `npm: enabled: true` to the `brunch-config.coffee` and installing e.g. jquery as dependency in your `package.json`. The build will fail if the `overrides` object is missing, but succeeded with my changes.

PR #1003, opened 2015-08-05
Only listen to stdin if --stdin flag is given

### LINKED ISSUE #998: Re-enable brunch to run in background
Issue #922 caused brunch to no longer run in the background. We used to run it and be able to press Ctrl+Z and bg to have it run in the background. After #922, we see "stopped" and brunch refuses to run in the background. If we back out #922, everything is cool.

Is there a quick fix for this?

josevalim:
Closes #998

I would really appreciate if there was a release so we can update Phoenix to use the `--stdin` flag from now on.

Also, I did not make the flag hidden, because folks would see it and wonder what it is about. I tried to make it as descriptive as possible though.

Thank you!

### DIFF
--- a/src/cli.coffee
+++ b/src/cli.coffee
@@ -35,6 +35,7 @@ watchCmd = program
   .option('-p, --port [port]', 'if a `server` option was specified, define on which port
  the server would run')
   .option('-d, --debug', 'print verbose debug output to stdout')
+  .option('--stdin', 'listen to stdin and exit when stdin closes')
   .action(commands.watch)
 
 addDeprecatedOpts = ->
--- a/src/watch.coffee
+++ b/src/watch.coffee
@@ -47,6 +47,7 @@ generateParams = (persistent, options) ->
   if options.production? or options.optimize?
     params.env.unshift 'production'
   params.persistent = persistent
+  params.stdin = options.stdin?
   if options.publicPath
     params.paths = {}
     params.paths.public = options.publicPath
@@ -499,7 +500,7 @@ bindWatcherEvents = (config, fileList, compilers, linters, watcher, reload, onCh
     onChange()
     changeFileList compilers, linters, fileList, path, false
 
-  if config.persistent
+  if config.persistent and config.stdin
     process.stdin.on 'end', -> process.exit(0)
     process.stdin.resume()
 


src/cli.coffee
```diff
@@ -35,6 +35,7 @@ watchCmd = program
   .option('-p, --port [port]', 'if a `server` option was specified, define on which port
  the server would run')
   .option('-d, --debug', 'print verbose debug output to stdout')
+  .option('-l, --stdin', 'listen to stdin and shut down when stdin closes')
```
es128: what about the idea of the shorthand being just `-` to follow the convention of other unix tools?

shreeve: @es128 - It seems like the `-` option works fine, but then I saw this elsewhere:

```
10:51 AM shreeve@ice ~/Data/Work/Code/xo-next> bws -
Error: ENOENT, no such file or directory
  at Error (native)
  at process.chdir (/Users/shreeve/Data/Work/Code/brunch/node_modules/chokidar/node_modules/readdirp/node_modules/graceful-fs/polyfills.js:18:9)
  at module.exports.watch (/Users/shreeve/Data/Work/Code/brunch/src/watch.coffee:587:15)
  at start (/Users/shreeve/Data/Work/Code/brunch/src/index.coffee:9:3)
  at Command.listener (/Users/shreeve/Data/Work/Code/brunch/node_modules/commander/index.js:289:8)
  at Command.emit (events.js:110:17)
  ...
```

Is this easy to cleanup? If so, then the `-` option seems preferable to `-l`.

es128: Yes that's probably easy to clean up at https://github.com/brunch/brunch/blob/master/src/watch.coffee#L585-L586

Either just wrap in a `try` or specifically exclude `-`

shreeve: I guess `commander` thinks the `-` option is just a leftover argument, so it gets treated differently and does *not* act the same as passing `--stdin`. Perhaps using the `-l` short option would work best. Unless you can poach the `-` off and make it act like `--stdin`.

josevalim: I decided to not use "-" because it usually means you can write stuff in
the stdin and it will be read. Here we just care about EOF.



-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

es128: oh ok, I had taken your word for it before that it would work. So I suppose we should just leave it like this, or just remove the shorthand flag.

@josevalim why did you choose `-l`?

josevalim: I choose -l for "listen" but feel free to remove it. I am not attached to
it. :)

-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

es128: It's your PR :smile:

I don't have strong feelings about it, but tempted to say I'd prefer leaving it at just the explicit `--stdin`. It's not going to be manually typed in enough (or ever) to justify a shorthand letter.

paulmillr: + to es128

josevalim:
Force pushed. This is ready to go!

josevalim:
Please do a release when you can so I can update Phoenix. :)

es128:
https://github.com/brunch/brunch/commit/b8b25bc2c0e6dd21a8ad3d2a8e0157e770f3d3d2

josevalim:
Thank you! I appreciate it! :heart:

PR #1005, opened 2015-08-05
enable --stdin option

shreeve:
If this option is passed, a persistent brunch process will exit if stdin is closed.


es128:
So, by the looks of it, `stdin: true` could be added to the brunch-config also to make it unnecessary to add `--stdin` on the CLI each time. Should we add that to the config docs?

es128:
Wait, @josevalim already beat you to it in #1003

PR #1010, opened 2015-08-19
Badges in readme.md

PeterDaveHello:


es128:
Thanks, but I don't think these are adding any value here. Brunch does not currently have a robust test suite, and considering it is a tool, not a module people depend upon in their production apps, I don't think the info provided by the other badges would be useful at this point.

PeterDaveHello:
You can easily know that there is security issue in dependencies ...

[devDependency Status](https://david-dm.org/brunch/brunch#info=devDependencies)

es128:
> it is a tool, not a module people depend upon in their production apps

PeterDaveHello:
Fine.

Issue #1024, opened 2015-10-02
Brunch should come with a init command to init new projects with the dead-simple skeleton, out of the box


marcolz:
That is the only skeleton I really use, as most available skeletons these days are out of date, or never fully suiting me needs.

I'm pretty sure I'm not the only one to always use dead-simple and then start adding custom libraries.

Just an idea.

es128:
You're asking for `brunch init` as an alias for `brunch new brunch/dead-simple`?

marcolz:
Correct. ☺️

vendethiel:
Yeah, I know I'm contributing to this trend... That's also the issue with technologies that come by and go away so fast... Stuff gets outdated really, really rapidly...

paulmillr:
Agree.

es128:
It would be good if we could find a place to hold global settings, so users could pick whatever they want their default skeleton to be for the `init` command.

paulmillr:
The default one is dead-simple since 2.0, closing

PR #1025, opened 2015-10-06
Add init command

marcolz:
Add a new `init` command to quickly create new projects with a default skeleton (defined in the ENV, or 'dead-simple' otherwise).

```sh
brunch init ./fun-project
```

```sh
BRUNCH_INIT_SKELETON=dead-simpler brunch init ./funner-project
```

Towards #1024

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -3,6 +3,7 @@
 const program = require('commander');
 const commands = require('./');
 const aliases = {
+  i: 'init',
   n: 'new',
   b: 'build',
   w: 'watch'
@@ -14,6 +15,10 @@ Object.keys(aliases).forEach(key => {
 });
 
 program.version(require('../package.json').version).usage('[command] [options]');
+
+program.command('init [path]').description('Create new dead simple brunch project in path [.]. Short-cut: i').action(() => {
+  return commands["init"](program.args[0]);
+});
 
 program.command('new [skeleton] [path]').description('Create new brunch project in path [.]. Short-cut: n').action(() => {
   return commands["new"](program.args[0], program.args[1]);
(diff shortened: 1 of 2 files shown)


es128:
LGTM :+1:

marcolz:
Just realized my email address was out-of-date. Push-forced the amended commit.

es128:
Glad I fought back the urge to just hit the merge button, then

marcolz:
:+1: It's good to go now.

es128:
Leaving it for @paulmillr to review and do the honors

paulmillr:
Why not add it to `new` by default?

paulmillr:
@es128 hey?

es128:
@paulmillr you agreed with the concept in the original issue, and I'm not totally sure what you mean now. How would this be done on `brunch new`? 

Like `brunch new myproject` would use the default skeleton? And it would know that's what you meant compared with the existing usage of `brunch new` because `myproject` doesn't start with `gh:` or `git:\\`?

paulmillr:
Yes. Small breaking change for 2.0. `brunch new` is used much less often than `build` etc.

```
brunch new # Help
brunch new project # Create a dir `project` with default skeleton
brunch new .           # Init default skeleton in the current dir
brunch new project -s <uri>
brunch new project —skeleton <uri>
brunch new -s <url>
```

paulmillr:
Still thinking on the most common case. I just think that having two commands instead of one (when the second one is basically the first one with an argument) is confusing?

paulmillr:
```
# Current
brunch new <skeleton> <path>
brunch new simple
brunch new simple .
brunch new react project
```

es128:
I agree with the notion that I can think of better future enhancements that do something completely different with `brunch init`

So, we should probably close this, and redirect the effort toward overhauling `brunch new`

PR #1032, opened 2015-10-25
Promises for brunch

ghostd:
Hi,

Here is a patch to replace (most of) callbacks by promises.

I only tested it against a small project (3 js, 2 scss). this patch needs to be tested with an actual project (especially to check the performance).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+; http://editorconfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+
+[*.md]
+indent_size = 4
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 node_modules
 npm-debug.log
 .c9/
+.idea/
--- a/lib/fs_utils/asset.js
+++ b/lib/fs_utils/asset.js
@@ -47,17 +47,22 @@ class Asset {
   }
 
   /* Copy file to public directory. */
-  copy(callback) {
-    common.copy(this.path, this.destinationPath, (err) => {
-      let error;
-      if (err != null) {
-        error = new Error(err);
-        error.code = 'Copying';
-      }
+  copy() {
+    const saveTime = (p) => {
       this.copyTime = Date.now();
-      this.error = error;
-      callback(error);
-    });
+      return p;
+    };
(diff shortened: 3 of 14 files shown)


paulmillr:
That's a good one. Thanks Vincent! I'll test it on a real project.

paulmillr:
My initial suggestion would be to move away from bluebird to native promises.

ghostd:
Here is a version without bluebird

Since i removed 'async-each' on this branch, i think we'll need to add a 3rd party dependency to get back some performance. (check around my 'Array.reduce' and 'Promise.all' calls in the code)

paulmillr:
It seems like we have quite a lot cases where we simply do `reject` if there's an async error and `resolve` if no error. Could you please make a util for this purpose like `promisify` in bluebird?

```javascript
const promisify = (fn) => {
  return function() {
    return new Promise((resolve, reject) =>
      const args = [].slice.call(arguments);
      args.push((error, data) => {
        if (error != null) return reject(error);
        resolve(data);
      });
      fn.apply(null, args);
    });
  };
};

const read = promisify(fs.readFile);
```

paulmillr:
Thanks — will check this tomorrow.

paulmillr:
Doesn't work with a project.

```
30 Oct 22:04:44 - error: [TypeError: components.sort is not a function]
```

Seeing this in the console.

paulmillr:
There seem to be no speed penalty while using this branch. Interesting.

paulmillr:
Last-minute change: Could you please squash all commits into one? I think we're good to merge this.

@es128 ?

es128:
+0 meh

https://twitter.com/henrikjoreteg/status/661053038690340865

I don't necessarily agree with that sentiment 100%, but in the context of this PR and what motivated it I do.

paulmillr:
Totally agree, there is no simplification here.

paulmillr:
ideally there would just be `async` / `await` and `yield` shit, don't know how far away we are from there.

ghostd:
i agree these commits does not simplify the code.
So do we drop this PR ?

paulmillr:
No. Let's keep it around. Could you squash all commits into one please?

paulmillr:
gh-1036

paulmillr:
@ghostd thanks, it's in master now.

PR #1034, opened 2015-10-27
config.md - default ignore pattern

Bartmax:
Add instructions to add ignore conventions, including the default ignore
pattern.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -86,6 +86,19 @@ conventions:
   assets: /files[\\/]/  # vendor/jquery/files/jq.img
 ```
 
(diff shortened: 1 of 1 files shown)


docs/config.md
```diff
@@ -86,6 +86,18 @@ conventions:
   assets: /files[\\/]/  # vendor/jquery/files/jq.img
 ```
 
+If you want to add to the ignore pattern instead of replace, you must copy the defaults into your config.
+
+Default ignore pattern:
+```
```
es128: specify coffeescript here for the syntax highlighting

PR #1047, opened 2015-11-26
Switch the -p and -P flag documentation

ryansobol:
Looks like these doc changes slipped through the cracks. :memo:


es128:
Thanks!

Issue #1045, opened 2015-11-23
Post onCompile string replaces not working


mcfarljw:
I just noticed that post onCompile string replaces aren't working consistently using version `2.0.0` and `2.0.1`. The replace doesn't work at all when build is run. When the initial watch is started replaces don't happen either, but it *sometimes* replaces things when files are changes during the watch.

When I revert back to `1.8.5` post string replaces work properly.

```
//requirements
var shelljs = require('shelljs');

//exports.config
onCompile: function(files) {
    files.map(function(file) {
        return file.path;
    }).forEach(function(path) {
        shelljs.sed('-i', '{!application-title!}', 'My Application', path);
    });
}
```

PR #1050, opened 2015-12-04
We have to return the 'write' promise, else the callbacks are called …

ghostd:
…before the files are actually written


ghostd:
This fixes #1045

es128:
LGTM, and thanks @ghostd for working on fixing these bugs. How do you feel about the idea of also writing unit tests for these situations to make sure we don't have regression problems with the same mistakes in the future?

Issue #1038, opened 2015-11-15
NPM: performance with large packages


milgner:
Brunch: 1.8.5
Node: v3.0.0
Npm: 2.13.3

First the initial problem:

Working on a [Phoenix](http://phoenixframework.org/) project, after adding `material-ui` to my `package.json`, brunch stopped working. I first increased the `nofile` ulimit to 100.000, which then resulted in the error:

```
$ node node_modules/brunch/bin/brunch build
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory
Aborted (core dumped)
```

Not sure how to go about addressing this. My brunch config is still unchanged from the [default template](https://github.com/phoenixframework/phoenix/blob/master/installer/templates/static/brunch/brunch-config.js).

Sure, the `material-ui` package is quite large, containing 3866 files alone, but it should still be possible to use?

It may be part of the broader issue in #1023 but I'm willing to try anything; let me know if you need additional info.

paulmillr:
I would suggest to use standalone files for now. Performance may definitely be a concern until we rework it.

milgner:
For now I have switched to webpack to get things working, but I wanted to notify you about the problem with large NPM modules. Brunch certainly feels great, keep up the good work!

PR #1059, opened 2015-12-11
NPM integration for front-end packages

goshakkk:
This improves upon deppack. #1023 

Some highlights:

* every used file from each package gets wrapped into a module (like "react/lib/Object.assign") — previously deppack exposed only index.js as "package-name"
* respecting `browser` property from package.json to require different files in the browser environment
* `process` shim to make some things work
* require tree is only traversed during watcher setup, making succ build significantly faster
* it is possibly to use direct dependencies' stylesheets files
* adding `material-ui` (or any other large package) doesn't break the build (#1038)

Example usage:

```coffee
exports.config =
  npm:
    enabled: true
    packages: ['react', 'react-dom', 'react-router']
```

you can also use stylesheets & expose globals

```coffee
exports.config =
  npm:
    enabled: true
    packages: ['react', 'react-dom', 'react-router', 'pikaday']
    styles:
      pikaday: ['css/pikaday.css']
    globals:
      Pikaday: 'pikaday'
```

Some further ideas/considerations:

* do we need to properly shim things? other than `process`, most front-end packages seem to work just fine with `{}` shims for node core modules

* instead of specifying the front-end packages explicitly, we might at some point figure these out based on application code. It's probably going to be too complex for this first iteration, though

* do we need some special handling for other types of assets coming from the packages, like images or fonts?

* is inlining deppack into brunch a way to go, or should it be modularized? At this point, it's not like the public API for deppack is completely finished so it might make sense to inline it, for the time being

* `read-components` is doing bower/components integration. Should deppack be somehow reconciled with it? They serve similar purpose, although their approaches and specifics are different

* finally, this is for CommonJS-style only

* the docs should probably be updated, and an example project for this be created

### DIFF
--- a/lib/deppack.js
+++ b/lib/deppack.js
@@ -1,0 +1,282 @@
+'use strict';
+const sysPath = require('path');
+const fs = require('fs');
+const detective = require('detective');
+const browserResolve = require('browser-resolve');
+const requireDefinition = require('commonjs-require-definition');
+const each = require('async-each');
+
+const specialShims = { process: { env: { NODE_ENV: process.env.NODE_ENV } } };
+const emptyShims = [
+  'assert', 'buffer', 'child_process', 'cluster', 'crypto', 'dgram', 'dns',
+  'events', 'fs', 'http', 'https', 'net', 'os', 'path', 'punycode',
+  'querystring', 'readline', 'repl', 'string_decoder', 'tls', 'tty', 'url',
+  'util', 'vm', 'zlib'
+];
+
+const shims = Object.assign(emptyShims.reduce((memo, el) => {
+  memo[el] = {};
+  return memo;
+}, {}), specialShims);
+
+const id = x => x;
+const not = f => i => !f(i);
+
+const uniq = list => {
+  return Object.keys(list.reduce(function(obj, _) {
+    if (!obj[_]) obj[_] = true;
+    return obj;
+  }, {}));
+};
+
+const readFile = (path, callback) => {
+  return fs.readFile(path, {encoding: 'utf8'}, callback);
+};
+
+const getModuleRootPath = path => {
+  const split = path.split('/');
+  const index = split.lastIndexOf('node_modules');
+  return split.slice(0, index + 2).join('/');
+};
+
+const getModuleRootName = path => {
+  const split = path.split('/');
+  const index = split.lastIndexOf('node_modules');
+  return split[index + 1];
+};
+
+const getModuleMainPath = path => sysPath.dirname(getMainFile(path));
+const packageJson = filePath => require(relativeToRoot(filePath, 'package.json'));
+const browserMappings = filePath => packageJson(filePath).browser || {};
+const relativeToRoot = (filePath, relPath) => sysPath.join(getModuleRootPath(filePath), relPath);
+
+const isRelative = path => path.slice(0, 2) === './';
+
+const globalBrowserMappings = filePath => {
+  const brMap = browserMappings(filePath);
+
+  return Object.keys(brMap).filter(not(isRelative)).reduce((newBrMap, key) => {
+    const val = brMap[key];
+    newBrMap[key] = generateModuleName(relativeToRoot(filePath, val));
+    return newBrMap;
+  }, {});
+};
+
+const expandedFilePath = filePath => {
+  const brMap = browserMappings(filePath);
+
+  Object.keys(brMap).filter(isRelative).forEach(key => {
+    const val = brMap[key];
+    if (filePath === relativeToRoot(filePath, val)) {
+      filePath = relativeToRoot(filePath, key);
+    }
+  });
+
+  return filePath;
+};
+
+const isMain = filePath => getMainFile(filePath) === filePath;
+
+const getNewHeader = (moduleName, source, filePath) => {
+  const brMap = globalBrowserMappings(filePath);
+
+  const r = isMain(filePath) ? "function(n) { return req(n.replace('./', '" + moduleName + "/')); }" : "req";
+
+  return `require.register('${moduleName}', function(exports,req,module){
+    var require = __makeRequire((${r}), ${JSON.stringify(brMap)});
+    ${source}
+  });`;
+};
+
+const generateModule = (filePath, source) => {
+  filePath = expandedFilePath(filePath);
+  const rootMod = getModuleRootName(filePath);
+  const mn = generateModuleName(filePath);
+
+  return getNewHeader(mn, source, filePath);
+};
+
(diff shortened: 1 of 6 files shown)


paulmillr:
1. This seems like a bad UX for me: `packages: ['react', 'react-dom', 'react-router']` — but yeah, ok for the first iteration
2. `do we need to properly shim things` — nope.
3. `do we need some special handling for other types of assets` — probably, later.
4. Let's inline deppack for now.
5. Yes, if it's simpler to integrate deppack with read-components then yes.
6. Commonjs ftw.
7. Could you update the docs too?

lib/deppack.js
```diff
@@ -0,0 +1,289 @@
...
+  'tls': {},
+  'tty': {},
+  'url': {},
+  'util': {},
+  'vm': {},
+  'zlib': {}
+};
+
+const id = x => x;
+const uniq = arr => arr.filter((i, pos) => arr.indexOf(i) === pos);
```
paulmillr: There's a room for optimization here. This is O(n²) or so.

lib/deppack.js
```diff
@@ -0,0 +1,289 @@
+'use strict';
+const sysPath = require('path');
+const fs = require('fs');
+const detective = require('detective');
+const browserResolve = require('browser-resolve');
+const requireDefinition = require('commonjs-require-definition');
+const each = require('async-each');
+
+const shims = {
+  'assert': {},
```
paulmillr: We can just make a list of items and assign an empty object to each of those. 30 less LOCs.

lib/deppack.js
```diff
@@ -0,0 +1,289 @@
...
+const relativeToRoot = (filePath, relPath) => sysPath.join(getModuleRootPath(filePath), relPath);
+
+const isRelative = path => path.slice(0, 2) === './';
+
+const globalBrowserMappings = filePath => {
+  const brMap = browserMappings(filePath);
+
+  const newBrMap = {};
+
+  Object.keys(brMap).filter(not(isRelative)).forEach(key => {
```
paulmillr: `.reduce`

lib/deppack.js
```diff
@@ -0,0 +1,289 @@
...
+  return newBrMap;
+};
+
+const expandedFilePath = filePath => {
+  const brMap = browserMappings(filePath);
+
+  Object.keys(brMap).filter(isRelative).forEach(key => {
+    const val = brMap[key];
+    if (filePath === relativeToRoot(filePath, val)) {
+      filePath = relativeToRoot(filePath, key);
```
paulmillr: We should probably stop the iteration here. Can we use `for-in & break`? Or, `Array#some`.

lib/deppack.js
```diff
@@ -0,0 +1,289 @@
...
+
+  return mn;
+};
+
+const _mainFile = (root, json) => {
+  const depMain = typeof json.browser === 'string' ? json.browser : json.main || 'index.js';
+  const fileOrDir = sysPath.join(root, depMain);
+  var isDir;
+  try { isDir = fs.lstatSync(fileOrDir).isDirectory() } catch(e) { isDir = false }
+  const main = isDir ? sysPath.join(fileOrDir, 'index.js') : fileOrDir.indexOf('.js') === -1 ? fileOrDir + '.js' : fileOrDir;
```
paulmillr: let's split the line in three or so.

lib/deppack.js
```diff
@@ -0,0 +1,289 @@
...
+  const mn = getModuleRootName(filePath) + (isMain(filePath) ? '' : filePath.replace(rp, '').replace('.js', ''));
+
+  return mn;
+};
+
+const _mainFile = (root, json) => {
+  const depMain = typeof json.browser === 'string' ? json.browser : json.main || 'index.js';
+  const fileOrDir = sysPath.join(root, depMain);
+  var isDir;
+  try { isDir = fs.lstatSync(fileOrDir).isDirectory() } catch(e) { isDir = false }
```
paulmillr: sync operations is a big "no". Brunch should always use async methods when possible.

lib/deppack.js
```diff
@@ -0,0 +1,289 @@
...
+  'util': {},
+  'vm': {},
+  'zlib': {}
+};
+
+const id = x => x;
+const uniq = arr => arr.filter((i, pos) => arr.indexOf(i) === pos);
+const not = f => i => !f(i);
+
+const readFile = (path, callback) => {
```
paulmillr: Could we use promises in this file?

lib/deppack.js
```diff
@@ -0,0 +1,289 @@
...
+
+  const readItem = (item, cb) => readFile(item.file, (err, data) => {
+    if (err) return cb(err);
+    cb(null, { file: item.file, data: data });
+  });
+
+  const detect = i => { return { file: i.file, deps: detective(i.data) }; };
+
+  const resolveItems = (items, cb) => {
+   each(items, readItem, (err, redItems) => {
```
paulmillr: Wrong indentation?

lib/deppack.js
```diff
@@ -0,0 +1,289 @@
...
+      }
+    };
+  `);
+  files.forEach(processor);
+  root.add('})();');
+
+  insertGlobals(config, root);
+};
+
+module.exports = loadFile;
```
paulmillr: Let's export it as `loadFile` or so too. Let's keep `module.exports` an object.

package.json
```diff
@@ -41,7 +40,12 @@
     "pushserve": "^1.0.0",
     "quickly-copy-file": "~0.1.0",
     "read-components": "~0.6.0",
-    "source-map": "~0.5.0"
+    "source-map": "~0.5.0",
+
+    "async-each": "~0.1.6",
+    "browser-resolve": "~1.7.2",
+    "commonjs-require-definition": "^0.2.0",
```
paulmillr: duplicate

lib/deppack.js
goshakkk: In this particular case, is it a big deal though? Having this _mainFile function as a sync one simplifies things quite a bit

junior-ales:
Hi,

I'm trying to use react as a regular node module in my project (requiring it like "var ReactDOM = require('react-dom');") but it's not working. It doesn't find the proper module.

The changes of this pull request are already working for the brunch version 2.0.4? Thanks for the support.

paulmillr:
@junior-ales this is not in yet.

goshakkk:
@junior-ales you can try it with master brunch. Btw, if you do, make sure you have `react-dom` in your package.json as well as `react`.

junior-ales:
great, thanks @paulmillr and @goshakkk

PR #1067, opened 2015-12-16
Add ESLint

goshakkk:
#1043 

I've left a few rules disabled:

* single quotes. It seems most of the codebase uses single quotes, but there are some instances (e.g. when reading paths from configs as `paths["public"]`) that still use double quotes. I'm wondering if this is in fact intended, or whether we should switch to single-quotes altogether?

* indentation, it's mostly consistent two spaces, however there are occasions when this rule is broken, notably when using promises:

```javascript
something.then(result => {
    // four spaces here!
    return doSomething(result);
  }, error => {
    logger.error(error);
  }
);
```

* prefer template strings over concatenation. There are roughly 70 instances of concatenation right now and I'm not sure whether to keep it or use the templates

Some rules also generate warnings currently without breaking the build:

* use of magic numbers, as to encourage giving them useful names
* comments with "TODO:", "FIXME:", etc
* non-camelcased identifiers


paulmillr:
need to fix merge conflicts

goshakkk:
Fixed. 

So wdyt regarding quotes/indentation/temp strings?

paulmillr:
1. Let's use single quotes.
2. Let's use two spaces always, promises should look like this:
    
    ```
    something.then(result => {
      return doSomething(result);
    }, error => {
      logger.error(error);
    });
   ```
3. Template strings are preferred for 2 >= variables

goshakkk:
Done, done, template strings rule is left disabled because we can't enforce 2+ vars here. Does it look good to you?

PR #1089, opened 2016-01-05
Handle shim requires during dep discovery

goshakkk:


paulmillr:
Travis seems to be breaking?

PR #1092, opened 2016-01-05
Ignore deps without main file

goshakkk:


lib/deppack.js
```diff
@@ -183,6 +184,12 @@ const load = (config, json, deps) => new Promise((resolve, reject) => {
     }
 
     const file = _mainFile(depPath, depJson);
+
+    if (!fs.existsSync(file)) {
```
paulmillr: async io is not quite cool

Issue #971, opened 2015-05-12
the watcher compiles a changed file as an empty file


maxlath:
I'm having this weird, exhausting issue since a few weeks.
The pattern seems to be this one: when I edit a file, around 1/10 of the times the file updated just before is compiled as empty.

Brunch shows no error, I got to detect the error from `ReferenceErrors` and alikes (when it's an empty js module) or missing DOM elements (when it's an empty handlebars template) that show up as the module is empty, getting me paranoid on every error: "is that coming from my code or from this random f*** bug?". I generally solve it by re-editing the empty file to force it to be recompiled.

The logs don't seem to show anyting (debug mode):
```
12 May 14:40:27 - info: compiled initialize.coffee and 328 cached files into app.js in 1271ms
  brunch:file-list Reading 'app/modules/comments/comments.coffee' +28s
  brunch:watch File 'app/modules/comments/comments.coffee' received event 'change' +0ms
  brunch:pipeline Compiling 'app/modules/comments/comments.coffee' with 'CoffeeScriptCompiler' +1ms
  brunch:helpers Wrapping '"modules/comments/comments"' with commonjs +20ms
  brunch:source-file Generated source map for 'app/modules/comments/comments.coffee'  +0ms
  brunch:file-list Compiled file 'app/modules/comments/comments.coffee'... +6ms
  brunch:write Writing 1/4 files +72ms
  brunch:generate Concatenating [some files], app/initialize.coffee, [LOTS of files] to public/javascripts/app.js +375ms
  brunch:common Writing file 'public/javascripts/app.js' +363ms
  brunch:common Writing file 'public/javascripts/app.js.map' +290ms
12 May 14:40:56 - info: compiled comments.coffee and 328 cached files into app.js in 1134ms
```
and as a result I get
```
;require.register("initialize", function(exports, require, module) {

});
```
instead of 
```
;require.register("initialize", function(exports, require, module) {
   [my module code here]
});
```

Here is my config file:
```
exports.config =
  notifications: yes
  sourceMaps: true
  files:
    javascripts:
      defaultExtension: "coffee"
      joinTo:
        'javascripts/vendor.js': /^(vendor\/scripts|bower_components)/
        'javascripts/app.js': /^app/
      order:
        before: [
          "bower_components/jquery/dist/jquery.js"
          "bower_components/underscore/underscore.js"
          "bower_components/backbone/backbone.js"
        ],
        after: [
          "vendor/*"
        ]

    stylesheets:
[...]

PR #1093, opened 2016-01-06
Small fix to watching for some editors

tprk77:
Some editors, such as Emacs, make backups while saving. This causes
files to be momentarily empty before the new contents are written. This
then results in a race between Brunch and the editor, trying to compile
the file and write the file at the same time. See also issue #971.

As a workaround, it's possible to delay the watch. This adds a new
watcher config option `fixEditorSave`. When true, the watcher will wait
up to 50ms for the write to stabilize.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -336,6 +336,7 @@ const normalizeConfig = config => {
   normalized.packageInfo = {};
   normalized.persistent = config.persistent;
   normalized.usePolling = !!(config.watcher && config.watcher.usePolling);
+  normalized.awaitWriteFinish = !!(config.watcher && config.watcher.awaitWriteFinish);
   config._normalized = normalized;
   ['on', 'off', 'only'].forEach(key => {
     if (typeof config.plugins[key] === 'string') {
--- a/lib/watch.js
+++ b/lib/watch.js
@@ -139,6 +139,8 @@ class BrunchWatcher {
     const config = this.config._normalized;
     const persistent = config.persistent;
     const usePolling = config.usePolling;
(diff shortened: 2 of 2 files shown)


tprk77:
Let me know if you want me to adjust/rework anything!

paulmillr:
Two ideas:

1. Can we rename it to a more generic name? `fix` in the name may sound to folks like it should be always used.
2. What if you want to enable it only for your environment, can this be a use case? Does it bring any disadvantages to other folks on your team?

tprk77:
I have no problems changing the name, do you have a suggestion? (I used `fix` only because it was shorter than `workaround`.) Maybe just expose `awaitWriteFinish` directly?

I think you're right that this is more related to the environment than the project. It would be cool if there were some kind of user specific config. I imagine other options could benefit from this as well, like `fileListInterval` and `usePolling`.

Anyway, I tried to make the delay pretty short (50ms), so this could almost always be turned on. So I think the impact is low. In my mind the performance penalty is definitely worth supporting more environments. Especially when the failure looks like a Brunch bug. At the same time, I know Brunch is very performance conscious, so I thought you might prefer this as a flag. Thoughts?

paulmillr:
Let's call the option `awaitWriteFinish: true / false`

tprk77:
Look any better? I wasn't sure if you meant for me to pass the flag directly to `chokidar` for the default `stabilityThreshold`/`pollInterval`. It's just that 2 seconds seems like a long time.

paulmillr:
Thanks dude!

Issue #1090, opened 2016-01-05
brunch build on new project hangs


chrismcg:
Running brunch build on a new project generated from the template builds but doesn't exit.

Brunch 2.1.1, Node 5.3.0, OS X 10.9.5 (Also happening on CI server with Ubuntu)

To reproduce:

```
$ npm install -g brunch
$ brunch new tb
$ cd tb
$ brunch build
```

I see:

```
05 Jan 08:58:54 - info: copied README.md in 262ms
```

However the build never terminates so any CI or production asset compilation hangs.

chrismcg:
Same if I do:

```
$ npm install -g brunch/brunch
```

To get latest master

chrismcg:
In my actual app I tried 2.1.0, 2.0.0, which still had the same problem. 1.8.5 terminates correctly.

paulmillr:
Could you please paste the output of `brunch build -d`?

chrismcg:
```
chris@gairmi % brunch build -d
  brunch:speed Created BrunchWatcher 428ms +0ms
  brunch:speed Loaded config 458ms +30ms
  brunch:speed Loading plugins 460ms +2ms
  brunch:speed Loaded plugin javascript-brunch 485ms +25ms
  brunch:speed Loaded plugin css-brunch 487ms +2ms
  brunch:speed Loaded plugin uglify-js-brunch 566ms +79ms
  brunch:speed Loaded plugin clean-css-brunch 673ms +107ms
  brunch:speed Loaded plugin auto-reload-brunch 710ms +37ms
  brunch:plugins Loaded plugins: javascript-brunch, css-brunch, auto-reload-brunch +2ms
  brunch:speed Loaded plugins 712ms +0ms
  brunch:speed Loaded watcher 713ms +1ms
  brunch:watch add package.json +28ms
  brunch:watch add bower.json +2ms
  brunch:watch add brunch-config.coffee +0ms
  brunch:watch add app/README.md +11ms
  brunch:list Reading app/README.md +1ms
  brunch:speed Starting compilation 756ms +1ms
  brunch:watch add app/assets/README.md +3ms
  brunch:asset Init app/assets/README.md directory=app/assets/ relativePath=README.md destinationPath=public/README.md +3ms
  brunch:speed Watcher is ready 764ms +2ms
  brunch:asset Copied app/assets/README.md +7ms
  brunch:write Writing 0/0 files +68ms
05 Jan 13:11:58 - info: copied README.md in 409ms
```

[...]

PR #1099, opened 2016-01-07
Fix build hangup

goshakkk:
brunch/brunch#1090

### DIFF
--- a/lib/ui.js
+++ b/lib/ui.js
@@ -1,13 +1,9 @@
 'use strict';
-const charm = require('charm')(process);
 const logger = require('loggy');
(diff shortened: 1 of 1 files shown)


lib/ui.js
```diff
@@ -24,6 +20,11 @@ const animatedProgress = (level, base) => {
     process.stdout.write(line);
   };
 
+  process.on('exit', () => {
```
paulmillr: We also need to unbind previous handlers before binding new ones.

goshakkk: Updated

PR #1103, opened 2016-01-09
Properly split path on Windows

josevalim:
See report here: https://github.com/phoenixframework/phoenix/issues/1463#issuecomment-170171217

Once the user reports it works on Windows, a 2.1.3 release
would be appreciated. I will drop a note here. :)


paulmillr:
213 is up

PR #1104, opened 2016-01-09
Smarter NPM integration

paulmillr:


package.json
```diff
@@ -47,7 +47,8 @@
     "charm": "1.0.0",
     "async-each": "~0.1.6",
     "browser-resolve": "~1.7.2",
-    "detective": "~4.0.0"
+    "detective": "~4.0.0",
+    "glob": "*"
```
paulmillr: That may break the build some time in the future. No stars please.

goshakkk:
This already delivers ~2x speed boost on one of our apps, compared to the previous deppack. Additionally, it's smarter, meaning you no longer have to specify in the config which packages to load.

Specifically for this PR, the following is still needed:

- [ ] investigate opportunities to speed up even further; and
- [ ] update docs

This revision has some of the following drawbacks, which should probably be addressed separately:

If a `brunch watch` is started with app requiring `a` and `b`, and then a `require` call for `b` is removed in the editor, then, even if nothing else references `b`, it will still be in the bundle. (Making the dev bundle bigger than it needs to be is about the only consequence.)

This is not of a big concern though, because a production bundle will be built separately via a `brunch build` invocation.

paulmillr:
1. Let's squash commits.
2. Let's remove hardcoded things from all places (e.g. `app` dir is defined by Brunches convention, same with `vendor`)
3. After thorough testing let's release 2.2 and start working on 2.3 with `watch` improvements for NPM.

lib/deppack.js
```diff
@@ -324,7 +267,99 @@ const processFiles = (config, root, files, processor) => {
...
+
+  return curr.join(',') === mods.join(',');
+};
+
+let modMap;
+
+const buildModMap = () => {
+  // this is needed to not throw on app-wise requires
+  // note that fileList.files can't be used because it's not fully populated until the first compilation
+  var mods = require('glob').sync('app/**/*');
```
paulmillr: Synchronous IO is not acceptable in Brunch. Too slow on big apps!

lib/deppack.js
```diff
@@ -324,7 +267,99 @@ const processFiles = (config, root, files, processor) => {
...
+    return map;
+  }, {});
+
+  modMap = Object.assign({}, modMap0, shimModMap);
+};
+
+const isJs = (fileList, path) => fileList.find(path).type === 'javascript';
+
+const isVendor = path => path.indexOf('vendor') === 0;
+const isApp = path => path.indexOf('app') === 0;
```
paulmillr: See config conventions / paths

lib/deppack.js
```diff
@@ -324,7 +267,99 @@ const processFiles = (config, root, files, processor) => {
...
+            fileList.watcher.changeFileList(p, false);
+          }
+        });
+        resolve(x);
+      });
+    });
+  };
+};
+
+module.exports.exploreDeps = exploreDeps;
```
paulmillr: can simply be `exports.exporeDeps`

paulmillr:
This needs a rebase.

goshakkk:
Rebased

lib/deppack.js
```diff
@@ -323,7 +251,105 @@ const processFiles = (config, root, files, processor) => {
...
+let modMap;
+
+const buildModMap = () => {
+  if (modMap) return Promise.resolve();
+
+  // this is needed to not throw on app-wise requires
+  // note that fileList.files can't be used because it's not fully populated until the first compilation
+  return promisify(glob)('app/**/*').then(mods => {
+    const modMap0 = mods.reduce((map, mod) => {
+      const name = mod.split('.').slice(0, -1).join('.').replace(/^app\//, '');
```
paulmillr: Still needs to match the proper convention.

lib/deppack.js
```diff
@@ -323,7 +253,105 @@ const processFiles = (config, root, files, processor) => {
...
+};
+
+let modMap;
+
+const buildModMap = () => {
+  if (modMap) return Promise.resolve();
+
+  // this is needed to not throw on app-wise requires
+  // note that fileList.files can't be used because it's not fully populated until the first compilation
+  return promisify(glob)('app/**/*').then(mods => {
```
paulmillr: и вот тут `${config.paths.watched}/**/*`, я так полагаю.

paulmillr: Also, don't forget about `assets` and `ignored` files.

docs/config.md
```diff
@@ -104,19 +104,14 @@ conventions:
 `Object`: configures NPM integration for front-end packages. Make sure you also declare the packages you depend on in your package.json `dependencies` section.
 
 * `npm.enabled`: `Boolean`: a toggle of whether the integration is enabled, defaults to `false`.
-* `npm.whitelist`: `Array`: a list of npm packages that should be included in the build.
-* `npm.blacklist`: `Array`: a list of npm packages that should not be included in the build.
 * `npm.globals`: `Object`: a mapping from global name (as a key) to the corresponding module name (string) to expose.
 * `npm.styles`: `Object`: a mapping from package name (string) to an array of stylesheet paths (relative to package root) to be included into the build.
 
-Use either `npm.packages` to explicitly whitelist, or `npm.exclude` to explicitly blacklist, or omit both to include everything.
```
paulmillr: Can we remove the docs later? They're still used by 2.1 folks

goshakkk: Well, so what about the website points to the docs from v2.1 tag? Docs in the repo should go in hand with the code I think.

PR #1112, opened 2016-01-16
Update plugin docs to use JS instead of Coffee

goshakkk:


paulmillr:
Let's switch to promises. @goshakkk

PR #1117, opened 2016-01-19
Npm deps resolving should also ignore test files

spyl94:


paulmillr:
do you have a specific use case from some package?

spyl94:
Sure, I'm using enzyme and chai, and my tests files are near my components files:

MyComponent-test.js
```
import React from 'react';
import {expect} from 'chai';
import {shallow} from 'enzyme';
import MyComponent from './MyComponent';

describe('<MyComponent />', () => {
  
it('should render', () => {
    const wrapper = shallow(<MyComponent />);
    expect(wrapper.find('li')).to.have.length(1);
  });
});
```
package.json
````
 "devDependencies": {
    "chai": "^3.4.1",
    "enzyme": "^2.0.0-rc1",
}
```
Because tests are only executed by mocha, it doesn't make sense for Brunch to search npm dependencies inside, imo.

paulmillr:
The NPM components don't export those files usually. We just may run into some issue when the file was ignored but it should not have been.

lib/deppack.js
```diff
@@ -301,6 +301,7 @@ const buildModMap = () => {
 };
 
 const isJs = (fileList, path) => fileList.find(path).type === 'javascript';
+const isTest = (path) => path.indexOf('-test') !== -1;
```
goshakkk: This looks very specific to your use case. Is it possible to use `conventions.ignored` for this? If not, what about adding a `config.npm.ignored` anymatch to tell deppack to ignore some app files?

spyl94:
I don't see any use case, do you have an example ? Why should brunch explore npm dependencies in files not compiled / delivered to the browser ?

Also those packages are not installed in production environment, it could break the compilation.

lib/deppack.js
spyl94: I found in the documentation of conventions: "Files named -test.<extension> are considered tests."
 
I don't see any use case where brunch need to explore npm dependencies in files not compiled / delivered to the browser. Also those packages are not installed in production environment, it could break the compilation.

But if you prefer, I'm ok with adding an `config.npm.ignored` pattern :)

paulmillr:
Package files are not usually require-ing tests, so after brunch 2.2 where we'll only include needed files this would be less useful

PR #1120, opened 2016-01-20
Update faq with npm integration information

VitaliyR:
Hi there.

Recently I've encountered a problem when I was need to override `main` attribute of some dependent  `package.json`.

Actually, I was need to grab only one .js file, but package `main` pointed to big compiled file.

After debugging for a while I've found, that brunch already supports overriding dependent packages manifests attributes not only for bower but for npm also.

I decided to update FAQ to help brunch's newcomers, maybe some of them facing it too.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -22,6 +22,56 @@ To add packages to your project:
 
 Example app with Bower integration: http://github.com/paulmillr/ostio
 
+## How to use NPM as client-side package manager?
+
+Brunch supports handling client-side dependencies using [NPM](https://npmjs.com) package manager.
+
+To do so, enable NPM integration in config. By default, brunch would look up for compatible node packages which can be used on client. You can tell branch to use explicit list of packages via `whitelist` attribute.
+
+Brunch can also handle styles of client-side libraries, but by providing `styles` attribute which is key-value object where key is package name and value is an array with relative to package path of styles which should be included.
+
+```coffeescript
+npm:
+    enabled: true
+    whitelist: [ 'react', 'react-dom', 'leaflet' ],
+    styles:
+        leaflet: [ 'dist/leaflet.css' ]
+files:
+    javascripts:
+        joinTo:
+            'js/vendor.js': /^node_modules/
+    styles:
+        joinTo:
+            'css/vendor.css': /^node_modules/
+```
+
+## How to handle other package assets?
+
+We're [working](https://github.com/brunch/brunch/issues/633) on it. As for now, you can solve it in different ways - by using `npm post-install` script, `onCompile` handler in config etc.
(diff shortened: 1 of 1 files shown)


paulmillr:
Thanks.

docs/faq.md
```diff
@@ -22,6 +22,56 @@ To add packages to your project:
 
 Example app with Bower integration: http://github.com/paulmillr/ostio
 
+## How to use NPM as client-side package manager?
+
+Brunch supports handling client-side dependencies using [NPM](https://npmjs.com) package manager.
+
+To do so, enable NPM integration in config. By default, brunch would look up for compatible node packages which can be used on client. You can tell branch to use explicit list of packages via `whitelist` attribute.
```
goshakkk: `whitelist` is gone in master (and so, will be gone in 2.2)

VitaliyR: I've read somewhere that 2.2 will predict which packages user is needed.
I found commits where info about whitelist & blacklist packages are removed. Too bad that we're updating docs before release - for instance, if I'm using latest published version I can't just to 'follow the docs'.

Should I remove from FAQ that information?

goshakkk: You should follow the docs at the tag not master?
On Thu, Jan 21, 2016 at 12:13 PM Vitaliy Ribachenko <
notifications@github.com> wrote:

> In docs/faq.md
> <https://github.com/brunch/brunch/pull/1120#discussion_r50375274>:
>
> > @@ -22,6 +22,56 @@ To add packages to your project:
> >
> >  Example app with Bower integration: http://github.com/paulmillr/ostio
> >
> > +## How to use NPM as client-side package manager?
> > +
> > +Brunch supports handling client-side dependencies using [NPM](https://npmjs.com) package manager.
> > +
> > +To do so, enable NPM integration in config. By default, brunch would look up for compatible node packages which can be used on client. You can tell branch to use explicit list of packages via `whitelist` attribute.
>
> I've read somewhere that 2.2 will predict which packages user is needed.
> I found commits where info about whitelist & blacklist packages are
> removed. Too bad that we're updating docs before release - for instance, if
> I'm using latest published version I can't just to 'follow the docs'.
>
> Should I remove from FAQ that information?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/brunch/brunch/pull/1120/files#r50375274>.
>
-- 
Cheers,
Gosha

VitaliyR: Then we need to update ./README.md with the link to latest released docs..? I mean to the docs in latest tag. 

About this one, I will update PR with newest information.

VitaliyR:
Moved to https://github.com/brunch/brunch/pull/1128

PR #1125, opened 2016-01-25
Update browser-resolve dependency.

colinbate:
The underlying `resolve` library has a bug in versions <1.1.7 which means it doesn't work on Windows when using relative paths (like `.`). This breaks the `npm.globals` config on Windows. I can confirm that updating this dependency does fix it.


goshakkk:
:+1: 

I was going to do this myself after https://github.com/defunctzombie/node-browser-resolve/issues/78. Thanks! 

@paulmillr merge this one

PR #1129, opened 2016-01-25
Override npm packages attributes fixes

VitaliyR:
Hello there.

After updating to 2.2 *(2.1.3 fine)* I've found that overrides section from my package.json not working anymore.

Overriding package.json is implemented in `./lib/deppack.js`.

I found that `browserResolve` doesn't care about overrides in my local `./package.json` because it is trying to read original package.json of module (`./node_modules/cool-package/package.json`).

### DIFF
--- a/lib/deppack.js
+++ b/lib/deppack.js
@@ -160,7 +160,7 @@ let paths;
 let nameCleaner;
 
 const getDepPackageJson = depPath => {
-  let depJson = require(sysPath.join(depPath, 'package.json'));
+  const depJson = require(sysPath.join(depPath, 'package.json'));
   applyPackageOverrides(depJson);
   return depJson;
 };
@@ -170,7 +170,7 @@ const globalPseudofile = '___globals___';
 const resolveErrorRe = /Cannot find module '(.+)' from '(.+)'/;
 
 const applyPackageOverrides = (pkg) => {
-  let pkgOverride = overrides[pkg.name];
+  const pkgOverride = overrides[pkg.name];
 
   if (pkgOverride) {
     pkg = deepExtend(pkg, pkgOverride);
--- a/lib/deppack.js
+++ b/lib/deppack.js
@@ -160,17 +160,25 @@ let paths;
 let nameCleaner;
 
 const getDepPackageJson = depPath => {
-  const dep = getModuleRootName(depPath);
   let depJson = require(sysPath.join(depPath, 'package.json'));
-  if (overrides[dep]) {
-    depJson = deepExtend(depJson, overrides[dep]);
-  }
+  applyPackageOverrides(depJson);
   return depJson;
 };
 
 const globalPseudofile = '___globals___';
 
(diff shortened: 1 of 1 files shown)


lib/deppack.js
```diff
@@ -183,9 +191,13 @@ const xBrowserResolve = (mod, opts, cb) => {
         err = isGlob ? `Could not load global module '${mod}'.` : `Could not load module '${mod}' from '${src}'.`;
 
         if (packages[topLevel] == null) {
-          err = err += ` Possible solution: add '${topLevel}' to package.json and \`npm install\`.`;
+          err += ` Possible solution: add '${topLevel}' to package.json and \`npm install\`.`;
         } else {
-          err = err += ` Possible solution: run \`npm install\`.`;
+          if (overrides[mod]) {
+            err += ` Possible solution: check your overrides in package.json.`;
```
goshakkk: Having something in overrides shouldn't automatically rule out the `npm install` tip.

VitaliyR: Yes, but it was before just a hint to run npm install. And we can get here only if we couldn't find a module required by user.

goshakkk: I'm not sure I'm following. Suppose I have react in my package.json and an override section for it **but** I forgot to `npm install`. This PR will print the unhelpful error message about checking my overrides

VitaliyR: Okay, but we need to support throwing both errors because if we would just left that previous one (run npm install) and in the same time having a bug in overrides section (e.g. some typo `main.jss`) it would be not clear.

VitaliyR: @goshakkk btw, you don't need to re-run npm install after overriding smth. 
What do you think about:
```javascript
err += ` Possible solution: ${overrides[mod] ? 'check your overrides in package.json; ' : ''}run \`npm install\`. `;
```

Or we can also try to check if main file of package is found in `applyPackageOverride` function (and throw error if file not found)

paulmillr:
Thanks!

goshakkk:
Still I think the 'npm install' suggestion should be there even if the
overrides are present. Imagine a coworker pulling and forgetting to npm
install. It'll suggest overrides which is completely unrelated. In case of
overrides, how about "Possible suggestions: 'npm install' and check your
overrides"?
On Tue, Jan 26, 2016 at 12:37 AM Paul Miller <notifications@github.com>
wrote:

> Thanks!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/brunch/brunch/pull/1129#issuecomment-174675521>.
>
-- 
Cheers,
Gosha

VitaliyR:
@goshakkk as I wrote before, we can solve it either updating that error message or do a temporary solution (until we find a better/already working solution):
```javascript
if (! fs.existsSync(mod + pkg.main) { // + support of pkg.browser
   err += ' check ur overrides section';
} else {
  err += ' run npm install';
}
```

PR #1131, opened 2016-01-26
Deppack shims

goshakkk:
This PR is missing a shim for `zlib`, because some of its deps relies on `require`ing json files and brunch/deppack don't currently handle that. Since `zlib` is probably the least used shim, it might be ok to leave it out for now.

Not sure, but could it make sense create a separate package just to depend on shim packages? Otherwise brunch's packages.json is going to be cluttered with that, and the purpose is not exactly obvious. I grouped these in a separate "section" in packages for now...

The PR still needs some cleaning up and rebasing, so don't merge just yet :)

### DIFF
--- a/lib/deppack.js
+++ b/lib/deppack.js
@@ -196,9 +196,9 @@ const xBrowserResolve = (mod, opts, cb) => {
           err += ` Possible solution: add '${topLevel}' to package.json and \`npm install\`.`;
         } else {
           if (overrides[mod]) {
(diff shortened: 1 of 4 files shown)


paulmillr:
This branch has conflicts that must be resolved @goshakkk

paulmillr:
I think creating a separate package for shims would make sense, yes.

paulmillr:
Created a repo: https://github.com/brunch/node-browser-modules

PR #1142, opened 2016-01-30
Make process shim path relative

### LINKED ISSUE #1141: Getting errors with new process shim
Somewhere between 2.2.1 and 2.2.2 something happened to cause my modules to have errors. I think there is some invalid code being generated around the npm modules, something do to with the `process` "module".

Uncaught Error: Cannot find module "process" from "/"

I see in my generated vendor.js file I'm getting this is being output near the bottom, just before the auto-reload stuff and the root module for this file. `Reflect` in this case is something I have registered as an `npm.globals`. This is one of the conspicuous diffs from when I generate from 2.2.1

```js
require.register('process/browser', function(exports,require,module) {
    module.exports = require('process');
  });process = require('process');})();window.Reflect = require('es7-reflect-metadata/dist/browser');// shim for using process in browser

var process = module.exports = {};
var queue = [];
var draining = false;
// ... more of what looks to be the process/browser implementation.
```

Let me know if I can help you with anything on this one.

colinbate:
This fixes #1141. Just need to make sure the shim path is made relative.

### DIFF
--- a/lib/deppack.js
+++ b/lib/deppack.js
@@ -477,7 +477,7 @@ const exploreDeps = fileList => {
         //const origPaths = getFileFiles(path);
         addFileFiles(path, relPaths);
         if (usedShims.length === 0) {
-          fileList.watcher.changeFileList(fileShims.process, false);
+          fileList.watcher.changeFileList(makeRelative(fileShims.process), false);
         }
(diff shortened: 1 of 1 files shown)


paulmillr:
Thanks Colin!

PR #1143, opened 2016-01-30
File-based module name only for stream, detect if 'process' is used

goshakkk:


lib/deppack.js
```diff
@@ -453,6 +456,7 @@ const exploreDeps = fileList => {
     if (!isJs(fileList, path) || !mediator.npmIsEnabled) return Promise.resolve(x);
 
     const allDeps = isVendor(path) ? [] : detective(x.compiled);
+    const usesProcess = isVendor(path) ? false : x.compiled.indexOf('process.') !== -1;
```
colinbate: I suspect this would yield false positives (matching `myprocess.whatever()`, which isn't the end of the world, but it also could fail to match something like `getEnv = proc => proc.env.NODE_ENV; getEnv(process);`. Perhaps a bit contrived. I know regexes are more expensive, but something like `/\bprocess\b/` might be better.

lib/deppack.js
```diff
@@ -142,19 +142,19 @@ const getNewHeader = (moduleName, source, filePath) => {
     'req';
 
   // core-util-is, used by the stream shims, relies on the Buffer global
-  const includeBuffer = moduleName === 'core-util-is';
-  const glob = includeBuffer ? `var Buffer = require('buffer');\n` : '';
+  const includeBuffer = source.indexOf(' Buffer.') !== -1 || source.indexOf('new Buffer') !== -1;
```
colinbate: Similar to my comment below this would miss: `if (Buffer.isBuffer(buf)) { }`

PR #1146, opened 2016-02-03
Added support for custom hostnames.

Velok:
Had some issues with remote access to the server. The server was listen on localhost. Now it is possible to change the hostname in the brunch-config file with `config.server.hostname`.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -12,6 +12,7 @@ const startServer = (config, callback) => {
   if (callback == null) callback = Function.prototype;
   const serverOpts = config.server || {};
   const port = parseInt(serverOpts.port, 10);
+  const host = config.server.hostname || 'localhost';
   const serverCommand = serverOpts.command;
   const serverPath = serverOpts.path;
   const publicPath = config.paths.public;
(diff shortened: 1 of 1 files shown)


goshakkk:
Probably a CLI param would work better for that, instead of config? The same as `port`

Velok:
With pushserve alone i know its possible to pass a host param. But i want to use `brunch watch` along with the `auto-reload-brunch` plugin.

lib/server.js
```diff
@@ -12,6 +12,7 @@ const startServer = (config, callback) => {
   if (callback == null) callback = Function.prototype;
   const serverOpts = config.server || {};
   const port = parseInt(serverOpts.port, 10);
+  const host = config.server.hostname || 'localhost';
```
goshakkk: What I mean is, can this be getting from `brunch s -h whatever` and not app config? There probably could be a way to access that from auto-reload-brunch... A thing like this, doesn't feel like it belongs into app config

Velok: I tried to find a way to pass a cli arg but i was not successful. Im new in node-coding this project and git and all this exciting stuff anyways. But my first PullRequest at all has passed.. yay! :-) Thanks @paulmillr !

paulmillr:
Thanks Pascal. This is really useful!

PR #1147, opened 2016-02-03
Properly support source maps with 2+ sources

goshakkk:
Which is needed to properly handle source maps for pre-processed stylesheets

### DIFF
--- a/lib/fs_utils/source_file.js
+++ b/lib/fs_utils/source_file.js
@@ -3,11 +3,14 @@ const debug = require('debug')('brunch:file');
 const pipeline = require('./pipeline').pipeline;
 const helpers = require('../helpers');
 const smap = require('source-map');
+const fcache = require('fcache');
 
 const prettify = helpers.prettify;
 const identityNode = helpers.identityNode;
 const replaceBackSlashes = helpers.replaceBackSlashes;
 const isWindows = helpers.isWindows;
+
+const readFile = helpers.promisify(fcache.readFile);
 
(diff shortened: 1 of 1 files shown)


lib/fs_utils/source_file.js
```diff
@@ -53,6 +54,15 @@ const updateMap = (path, compiled, wrapped, sourceMap) => {
   if (suffix) node.add(suffix);
   node.source = path;
   node.setSourceContent(path, wrapperContent);
+
+  // the supplied source map might contain more than one source file
+  if (sourceMap && sourceMap.sources) {
+    sourceMap.sources.forEach(path => {
+      const file = fileList.find(path);
+      const content = file && file.source || fs.readFileSync(path).toString();
```
goshakkk: sometimes a file won't be on the file list (e.g. if it's `_something.sass` and `_*` files are ignored) so it's probably ok to read these synchronously

paulmillr: We read those files with fcache already.

lib/fs_utils/source_file.js
```diff
@@ -53,7 +56,14 @@ const updateMap = (path, compiled, wrapped, sourceMap) => {
   if (suffix) node.add(suffix);
   node.source = path;
   node.setSourceContent(path, wrapperContent);
-  return node;
+
+  // the supplied source map might contain more than one source file
+  const addSource = path => readFile(path).then(content => {
```
goshakkk: Updated to use `fcache` for reads

Issue #1082, opened 2015-12-28
preCompile broken in 2.0.*


erlandsona:
I'm trying to use the preCompile key in my brunch-config.coffee but it only works with the syntax specified in the before and after-brunch plugins if I'm on brunch version 1.8.5... preCompile does not work with brunch@2.0.*

brunch-config.coffee
```Coffee
# using spawn instead of exec to hook the output of child_processes to the parent process so we can see output of child_process during execution instead of printed from a buffer held in memory.... Would love to use spawnSync but won't work with async tests that need to be run (karma, protractor)...

 {spawn} = require 'child_process'

afterBeforeBrunch = (commands) ->
  options = stdio: 'inherit'
  commands.map (command) ->
    spwndChild = spawn('bash', ['-c', command], options)

exports.config =
  overrides:
    production:
      preCompile: (end) ->
        afterBeforeBrunch([
          'echo Do something before compilation...'
          'echo ..................................'
          'echo'
          'echo'
          'echo Compiling now...'
        ])
        end()
      onCompile: (generatedFiles) ->
        afterBeforeBrunch([
          'echo'
          'echo ########################################################'
          'echo "Perhaps you\'d like to run something post compilation?"'
          'echo ########################################################'
          'echo'
          'npm test'
        ])
```

goshakkk:
Submitted https://github.com/brunch/brunch/pull/1153 to mitigate this

PR #1153, opened 2016-02-05
Make preCompile work

goshakkk:
Since the `preCompilers` array is passed to `Promise.all` in the watcher, the array has to be of promises. 

Currently it contains functions so no pre-compiler gets fired. This PR addresses that by:

* making `preCompilers` an array of promises
* allowing individual `preCompile` to accept one argument, a callback
* **or,** allowing individual `preCompile` to accept no arguments and return a promise
* **or,** allowing individual `preCompile` to accept no arguments and return nothing

#1082


paulmillr:
Broke Travis.

goshakkk:
Fixed

Issue #1151, opened 2016-02-04
node_modules not being treated as vendor/not being ignored by babel?


tmjoen:
First of all, thank you so much for a wonderful piece of software!

I'm having some trouble with my brunch setup. I have a JS library I've compiled that compiles just fine, and works just great on its own. As soon as I try to include it in my web app project (by pulling it in through node_modules) it gets weird.

If I copy the file to my `web/static/vendor` directory, it gets bundled just fine with my other vendor files. If I try to include it from `node_modules/brando/priv/static/js/brando.js` I get errors in brunch:

    error: Resolving deps of node_modules/brando/priv/static/js/brando.js failed. 
    Could not load module './accordion' from '/{{ path to my app }}/node_modules/brando/priv/static/js'. 
    Possible solution: add '.' to package.json and `npm install`.

This is with the default `vendor` setup in brunch, I also tried setting it explicitly:

    vendor: /(^bower_components|node_modules|vendor)[\\/]/,

And also explicitly ignoring it in the `babel` section:

    babel: {
      // Do not use ES6 compiler in vendor code
      ignore: [
        /^(web\/static\/vendor|bower_components|node_modules)/
      ]
    },

But to no avail, still the same error.

So my question is, how can I force brunch to not "parse" this file and treat
it like the files in `web/static/vendor` even if it lives in `node_modules`?

    $ brunch -v                                                                                                                                                        
    2.2.3

I have a created a project to demonstrate here:

https://github.com/tmjoen/brunch_test

In `brunch-config.js`, If I comment out the `node_modules/brando/..` entries in `joinTo` and `watch`, it works.

goshakkk:
There currently is no way to opt-out some files from the deppack processing. (babel options have nothing to do with that)

Do you think having an `ignore` option in `config.npm` would be helpful for cases like that?

Alternatively, brunch should be smarter and tell npm dependencies from included static JS in `joinTo`, would could be more in line with the brunch's mantra of simplicity. 

I'll probably work on that tomorrow

tmjoen:
Thank you! The ability to ignore sounds good to me.

tmjoen:
This isn't really related to the npm ignore bit, but I'm a bit unclear on why it works from any directory containing "vendor", and not a directory I've set to be a vendor from the `vendor:` key in `brunch-config.js` though.

If I create a directory `web/static/testing`, copy the offending files there and add `web/static/testing` to the `vendor:` key, I get the errors.



[...]

PR #1155, opened 2016-02-05
Allow to statically include files from npm packages

goshakkk:
#1151


goshakkk:
*don't merge just yet*

goshakkk:
Should be ok to merge now

PR #1157, opened 2016-02-06
Handle nested dependencies

goshakkk:
This is going to allow to use v1 of some library in the app code, while
retaining the ability for some other dependency to rely on v2.

It works by generating module names like this for nested modules:
"some_dep/node_modules/conflicting_dep". In require, it tries to prepend
that prefix. If the require fails, it drops it and tries to require just
"dep".

The drawbacks of this iteration are the try-catch approach that first tries to load current_module/node_modules/required_module and falls back to required_module otherwise, but that could as well be improved at a later time.

### DIFF
--- a/lib/deppack/modules.js
+++ b/lib/deppack/modules.js
@@ -20,6 +20,14 @@ const getModuleRootName = path => {
   return split[index + 1];
 };
 
+const getModuleFullRootName = path => {
+  const split = path.split(sysPath.sep);
+  const indexS = split.indexOf('node_modules');
+  const indexE = split.lastIndexOf('node_modules');
+  return split.slice(indexS + 1, indexE + 2).join('/');
+};
+
+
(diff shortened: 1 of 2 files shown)


lib/deppack/modules.js
```diff
@@ -20,6 +20,14 @@ const getModuleRootName = path => {
   return split[index + 1];
 };
 
+const getModuleFullRootName = path => {
+  const split = path.split(sysPath.sep);
+  const indexS = split.indexOf('node_modules');
+  const indexE = split.lastIndexOf('node_modules');
+  return split.slice(indexS + 1, indexE + 2).join('/');
```
paulmillr: Syspath.sep?

goshakkk: This is going to be a *module name*, not some kind of path. Module names are separated with `/`

Issue #1044, opened 2015-11-22
Brunch: Research how we can reduce the # of dependencies / installation time.


paulmillr:
Right now Brunch pulls 331 NPM modules which may take up to 3 minutes even on a good connection. This is very bad / long.

paulmillr:
Full list of deps now:

```
├─┬ anymatch@1.3.0 
│ ├── arrify@1.0.0 
│ └─┬ micromatch@2.3.2 
│   ├─┬ arr-diff@1.1.0 
│   │ ├── arr-flatten@1.0.1 
│   │ └── array-slice@0.2.3 
│   ├── array-unique@0.2.1 
│   ├─┬ braces@1.8.2 
│   │ ├─┬ expand-range@1.8.1 
│   │ │ └─┬ fill-range@2.2.2 
│   │ │   ├── is-number@1.1.2 
│   │ │   ├── isobject@1.0.2 
│   │ │   ├─┬ randomatic@1.1.3 
│   │ │   │ └─┬ is-number@2.0.2 
│   │ │   │   └── kind-of@1.1.0 
│   │ │   └── repeat-string@1.5.2 
│   │ ├── preserve@0.2.0 
│   │ └── repeat-element@1.1.2 
│   ├── expand-brackets@0.1.4 
│   ├─┬ extglob@0.3.1 
│   │ ├─┬ ansi-green@0.1.1 
│   │ │ └── ansi-wrap@0.1.0 
│   │ └── success-symbol@0.1.0 
│   ├── filename-regex@2.0.0 
│   ├── is-extglob@1.0.0 
│   ├─┬ kind-of@2.0.1 
│   │ └── is-buffer@1.1.0 
│   ├── lazy-cache@0.2.4 
│   ├── normalize-path@2.0.1 
│   ├─┬ object.omit@2.0.0 
│   │ ├─┬ for-own@0.1.3 
│   │ │ └── for-in@0.1.4 
│   │ └── is-extendable@0.1.1 
│   ├─┬ parse-glob@3.0.4 
│   │ ├── glob-base@0.3.0 
│   │ └── is-dotfile@1.0.2 
│   └─┬ regex-cache@0.4.2 
│     ├── is-equal-shallow@0.1.3 
│     └── is-primitive@2.0.0 
├─┬ anysort@1.0.0 
│ └─┬ anymatch@1.0.0 
│   └─┬ minimatch@1.0.0 
│     ├── lru-cache@2.7.0 
[...]

PR #1166, opened 2016-02-08
Resolve paths to make glob matching work reliably

istr:
To (kind of) address #1044 (which is really due npm completely broken and their mislead "nested deps" policy imnsvho), and to have a clean separation between tooling, sources, and builds, I am using a setup that breaks with ```brunch``` default conventions on purpose with separated directories for
- the brunch installation (e.g. ```~/brunch-installation```), not in git
- the project source (e.g. ```~/brunch-source```), mounted read-only in continuous integration setup (aka ```app``` as per brunch default convention)
- the project build target (e.g. ```~/brunch-build```), mounted read/write in continuous integration setup (aka ```public``` as per brunch default convention)

Using ```anymatch``` glob in this setup fails, because ```brunch``` does not resolve paths. This leads to relative paths in the build that could never be matched successfully using globs like ```joinTo: 'app.js': [ '**/app/**.js', '!**/app/ignoremeplease/*.js']```:

```
DEBUG="brunch:*" ~/brunch-inst/node_modules/brunch/bin/brunch b
[...]
  brunch:watch add /var/home/istr/brunch-source/app/application.js +3ms
  brunch:list Reading ../brunch-source/app/application.js +0ms
  brunch:file Init ../brunch-source/app/application.js: isntModule=false isWrapped=true +1ms
  brunch:pipeline Compiling ../brunch-source/app/application.js @ JavaScriptCompiler +2ms
  brunch:modules Wrapping brunch-source/app/application @ commonjs +5ms
  brunch:list Compiled ../brunch-source/app/application.js +1ms
[...]
```

To make that work, this PR ```resolve()```s the paths having the nice side-effect that logging shows the absolute path to the sources in question, so you don't need to figure out by resolving the paths in your mind.

```
DEBUG="brunch:pipeline" ~/brunch-inst/node_modules/brunch/bin/brunch b
  brunch:pipeline Compiling /home/istr/brunch-source/app/application.js @ JavaScriptCompiler +0ms
  brunch:pipeline Compiling /home/istr/brunch-source/app/styles/main.scss @ SassCompiler +16ms
  brunch:pipeline Dependencies /home/istr/brunch-source/app/styles/main.scss @ SassCompiler +4ms
08 Feb 10:59:17 - info: compiled 2 files into 2 files, copied index.html in 387ms
```

### DIFF
--- a/lib/fs_utils/file_list.js
+++ b/lib/fs_utils/file_list.js
@@ -2,6 +2,7 @@
 const debug = require('debug')('brunch:list');
 const EventEmitter = require('events').EventEmitter;
 const normalize = require('path').normalize;
+const resolve = require('path').resolve;
 const fcache = require('fcache');
 const Asset = require('./asset');
 const SourceFile = require('./source_file');
@@ -185,6 +186,7 @@ class FileList {
   }
 
   _change(path, compiler, linters, isHelper) {
+    path = resolve(path);
     const ignored = this.isIgnored(path);
     if (this.is('assets', path)) {
       if (!ignored) {
@@ -208,6 +210,7 @@ class FileList {
   }
 
   _unlink(path) {
+    path = resolve(path);
     const ignored = this.isIgnored(path);
     if (this.is('assets', path)) {
       if (!ignored) this.assets.splice(this.assets.indexOf(path), 1);


goshakkk:
This *is* going to break deppack though, as it relies on `fileList` paths in the events being relative. 
Additionally, it is common for anymatch regexps in both config and conventions to include `^` to indicate the start of the line, which, unsurprisingly, will also be broken, should we opt for full paths... right?

@paulmillr

istr:
@goshakkk You're right, using this simple approach will break deppack.

However, in many cases it is highly preferable to use a combination of glob includes / excludes (very clean, no escapes, readable) over regexps.

Closing, as this approach seems to be too "brute-force". The issue with basically broken glob support, however, remains.

PR #1170, opened 2016-02-09
Add support for .json files in node_modules

goshakkk:


lib/fs_utils/source_file.js
```diff
@@ -114,7 +115,8 @@ const makeCompiler = (path, cache, linters, compilers, wrap) => {
 class SourceFile {
   constructor(path, compilers, linters, wrapper, isHelper, isVendor, fileList) {
     this.fileList = fileList;
-    const type = compilers && compilers[0] && compilers[0].type;
+    // treat json files from node_modules as javascript
+    const type = compilers && compilers[0] && compilers[0].type || deppack.isNpm(path) && 'javascript';
```
goshakkk: not the best piece but then, not sure if there is a better way to let deppack influence file type...

lib/fs_utils/file_list.js
```diff
@@ -195,7 +195,8 @@ class FileList {
       debug(`Reading ${path}`);
       fcache.updateCache(path, error => {
         if (error) throw new Error(formatError('Reading', error));
-        if (!ignored && (compiler && compiler.length)) {
+        // .json files from node_modules should always be compiled
+        if (!ignored && (compiler && compiler.length) || deppack.isNpm(path)) {
```
goshakkk: same as below... json files from node modules won't have a compiler

paulmillr: Can we ensure the code path is executed only for json files?

lib/fs_utils/file_list.js
```diff
@@ -195,7 +195,8 @@ class FileList {
       debug(`Reading ${path}`);
       fcache.updateCache(path, error => {
         if (error) throw new Error(formatError('Reading', error));
-        if (!ignored && (compiler && compiler.length)) {
+        // .json files from node_modules should always be compiled
+        if (!ignored && (compiler && compiler.length) || deppack.isNpmJSON(path)) {
```
goshakkk: > Can we ensure the code path is executed only for json files?

by being more explicit about function name, like this?

Issue #1161, opened 2016-02-07
Test NPM integration backwards-compatibility.


paulmillr:
Want to release Brunch 2.3 in a bit; with NPM support enabled by default. It would be great to test whether the integration would break existing projects:

1. Create a project using bower and some server-side NPM packages (no client-side). Preferably for Brunch 2.0 or so.
2. Update to Brunch 2.2, set config.npm.enabled = true, make sure the app does not break.

My guess is that it should not, but we better test it.

goshakkk:
Looks good on a small app with React & Exim via bower

razor-x:
Something broke here involving `NODE_ENV === 'production'`  between 2.2.3 and 2.3.0 on my new  [React skeleton demo app](https://github.com/makenew/tasty-brunch/tree/demo).

Seems like on 2.2.3, when serving locally and on the production build, React / Redux gets packaged up in Dev mode. You can see the console message on the [live version](https://makenew.github.io/tasty-brunch/): 

```
"Download the React DevTools for a better development experience: https://fb.me/react-devtools" app-a43faf17.js:3:20739
You are currently using minified code outside of NODE_ENV === 'production'. This means that you are running a slower development build of Redux. You can use loose-envify (https://github.com/zertosh/loose-envify) for browserify or DefinePlugin for webpack (http://stackoverflow.com/questions/30030031) to ensure you have the correct code for your production build.
```

On 2.3.0, when serving locally (and also in production) I get the console error: `Uncaught TypeError: Cannot read property 'NODE_ENV' of undefinedReactPerf.measureMethods @ ReactPerf.js:41(anonymous function)`. The app fails to load.

Ideally, React / Redux would be used in Dev mode locally, and be packaged for production mode with `--production`.

PR #1174, opened 2016-02-13
Don't create aliases for shims with proper names

goshakkk:
Should fix the issue mentioned in https://github.com/brunch/brunch/issues/1161#issuecomment-183551321

cc @razor-x

The dev mode off in prod build is a nice-to-have and will be added later


mindreframer:
It really does! Thanks for such quick resolution! :+1:

Issue #1177, opened 2016-02-16
Custom server doesn´t work in recent version.


szathiago:
Custom server doesn´t work in recent version.
brunch-config.coffee
  server:
    command: "php -S 0.0.0.0:3000 -t public"

command: brunch watch -s
 error: Initialization error - ReferenceError: server is not defined

szathiago:
works perfectly on 2.2.3

goshakkk:
Thanks for the report, I'll take a look!

szathiago:
Thanks a bunch!

PR #1178, opened 2016-02-16
Fix handling of custom servers

goshakkk:
Closes GH-1177


lib/server.js
```diff
@@ -72,7 +72,7 @@ const startServer = (config, callback) => {
       child.kill();
       return typeof cb === 'function' ? cb() : undefined;
     };
-    serverCb();
+    setTimeout(serverCb, 0);
```
paulmillr: Second argument is not needed.

PR #1181, opened 2016-02-18
Added documentation for accessing server from other machines on local network

BernieSumption:
By default, `brunch watch --server` is only accessible on the same machine a the server process. Added some documentation explaining how to access it on other machines.


docs/config.md
```diff
@@ -248,6 +248,7 @@ The server script must export a function that starts your custom server, either
 If a custom server is not present, Brunch will use [pushserve](https://github.com/paulmillr/pushserve). If using your own, only `port` from the following options can be set from the config.
 
 * `port`: port on which server will run. Default: `3333`
+* `hostname`: hostname on which server will run. Default `localhost`, which only permits connections from the same computer. Set to `0.0.0.0` to permit access from other computers.
```
Florian-R: Maybe worth adding the same a note about the restriction on local network.

Issue #1186, opened 2016-02-19
Relative paths within node_modules are not always expanded correctly


jacwright:
### Description

Using moment.js with locales causes a JS error because the module path is incorrect.

```js
var moment = require('moment');
// Make sure this locale is compiled into the build
require('moment/locale/af');
// Causes error Error: Cannot find module "moment/moment" from "moment/locale/af" at require
```
This is because when the module loader attempts to expand it to an absolute path it doesn't work in this case. My guess is that the `moment.js` file is being compiled to `"moment"` because of the repo's `"main"` setting in `package.json`.

### Environment

1. Brunch: 2.3.2
2. Node: 5.6.0
3. NPM: 3.6.0
4. Operating system: Mac

### `package.json` contents
```json
{
  ...
  "dependencies": {
    "moment": "^2.11.2"
  },
  "devDependencies": {
    "auto-reload-brunch": "^2.0.0",
    "brunch": "^2.3.2",
    "clean-css-brunch": "~2.0.0",
    "css-brunch": "~2.0.0",
    "html-brunch": "^1.6.1",
    "javascript-brunch": "~2.0.0",
    "less-brunch": "~2.0.0",
    "uglify-js-brunch": "~2.0.0"
  }
}
```

### brunch config contents
```js
exports.config = {
  npm: {
    enabled: true
  },
  files: {
    javascripts: {
      joinTo: {
        'js/app.js': /^app|^node_modules/,
        'js/test.js': /^test/
      }
    }
}
```

jacwright:
Would aliases be a solution here? So `moment.js` would be the module `moment/moment` and aliased to `moment` because of the `main` directive?

goshakkk:
@jacwright it should be aliasing... let me check

goshakkk:
We used to create aliases for everything, but after the folks from phoenix asked us to not do that when not necessary (https://github.com/phoenixframework/phoenix/issues/1493#issuecomment-176734287), we only did it for a few shims

@paulmillr I think it should be ok to revert to always including file-based alias for now, and add some kind of smarter detection later on

jacwright:
I think I have to agree with pheonix about not aliasing the full filesystem path. It is not useful and puts user-specific info into the generated code.

I've coded up a more elegant aliasing solution in the above referenced PR. It will auto-alias files with extensions.

paulmillr:
@jacwright live in Brunch 2.5 now.

jacwright:
:+1:

PR #1189, opened 2016-02-20
Include file-based aliases for main files

goshakkk:
#1186


paulmillr:
Can we make a temp option that reverts the new behavior?

goshakkk:
@paulmillr #1191

PR #1196, opened 2016-02-22
Allow to specify again conventions.vendor as an anymatch set

Florian-R:
Per the docs, `conventions.vendor` can be set as an `anymatch` matcher but a small regression (comes from 0c1dec1dddff18664de40a52302d2d4814be6be9 it seems) made it Regex only.

This fix that by using anymatch again.

cc'ing @goshakkk just in case as he's the author of the aforementioned commit


goshakkk:
@paulmillr this looks good :+1:

PR #1199, opened 2016-02-24
Align function as a wrapper with other types

Florian-R:
Currently, when specifying `modules.wrapper` as a function, the `conventions.vendor` and `modules.nameCleaner` aren't respected.

This PR allow to :
- Not having to check if a file is a vendor one
- Not having to cleanup the module name manually


goshakkk:
@paulmillr this also looks good to merge

PR #1212, opened 2016-02-29
Update FAQ to prefer NPM integration

goshakkk:


docs/faq.md
```diff
@@ -11,7 +11,10 @@
```
goshakkk: Should the `scaffolt` reference still stay here or be removed?

paulmillr: Let's remove it.

PR #1217, opened 2016-03-02
Make relative requires from browser'd files work

goshakkk:
Based on report here http://stackoverflow.com/questions/35711331/brunch-issue-trying-to-bundle-socket-io-client

---

@paulmillr do you think we can move the extensions PRs forward?

### DIFF
--- a/lib/deppack/modules.js
+++ b/lib/deppack/modules.js
@@ -152,10 +152,6 @@ const isMain = filePath => getMainCached(filePath) === sysPath.resolve(expandedF
 const getNewHeader = (moduleName, source, filePath, origPath) => {
   const brMap = globalBrowserMappings(filePath);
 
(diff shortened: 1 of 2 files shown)


zecaptus:
Thanks for reporting my issue :)
Maybe you can handle another problem on this file.

I'll write an issue for this ;)

lib/deppack/modules.js
```diff
@@ -172,6 +168,11 @@ const getNewHeader = (moduleName, source, filePath, origPath) => {
         aliasDef(fbModuleNameUnexp, moduleName) :
         '';
 
+    console.log(origPath, getModuleRootPath(origPath));
```
paulmillr: .

goshakkk: Fixed

Issue #880, opened 2014-10-06
Optimize workers implementation


es128:
Use `child_process.spawn` with stdio pipes.

Stop using `.send` methods, especially for compilation results, [as they are synchronous/blocking](http://nodejs.org/api/child_process.html#child_process_child_send_message_sendhandle). And stop using `cluster`, as this isn't really what it is intended for, therefore adding unneeded bloat.

Also explore getting child processes launched earlier, before all the config normalization. Look for opportunities to reduce duplicated start-up work, etc

PR #1220, opened 2016-03-05
Simpler workers

goshakkk:
Previous workers implementation is overly complex and does not work on Brunch 2.x.

They worked by running full-featured brunch instance in every worker. Besides the added complexity, this approach also won't work with deppack, since it needs to store the dependency graphs of all files in one place.

This PR will introduce a simpler implementation of workers. All deppack caches and most of brunch logic will be run inside the main process. The workers are going to handle compilation/linting and optimization. The master will send them the data and they will just call the needed plugins on that, and send the result back.

Workers functionality is opt-in and requires you to pass a CLI option `-J <num>`. It *is* better than storing that in the config, because number of workers that yield the speed increase depends on each machine, it's number of cores, so it doesn't belong to the config.

On one of our apps that takes about 15s to compile on 2.4.2, and 25s to compile production (uglified) build on 2.4.2.

With 2 workers, on a dual-core macbook, the dev build is 12s and production one is 20s (so, **~20%** increase).

To reference #880:

* pipes would not provide much benefit compared to `[cp/process].send` (which is also async in node 5)
* it's not like any tangible amount of time is spent in IPC, maybe 200-400ms tops in the before-mentioned app (2-4%), so probably shouldn't be our priority when optimizing
* cluster is not used anymore

Reducing duplicate config load work (which takes maybe a second on that app) is not possible because:

1. node doesn't have threads and shared memory, so the children have to themselves process the config... **or**
2. the master process could send the loaded config to all of its children, after it's been loaded, but it can't, because the config is more than just JSON, it contains closed functions and such in the exports, so we can't just serialize it

The work to be done is to further look how we can achieve a close to 50% increase with 2 cores.

The available options would be to either:

* merge this, update the docs about this new functionality, mentioning it's *experimental*, and incrementally introduce performance optimizations (aka what we did with deppack); or
* wait until it's perfect

---

- [x] squash commits
- [x] properly terminate the process with children on `brunch build`
- [x] handle brunch restarts so that workers could be used for `watch`
- [x] remove remains of 1.8.x workers
- [x] handle errors in workers
- [x] update docs
- [x] make sure `onCompile` still works
- [x] get rid of a global (use a mediator?)
- [x] pass source map to optimizers
- [x] ensure production builds work as they should


lib/application.js
```diff
@@ -520,6 +520,19 @@ exports.loadConfig = (persistent, opts) => {
       .then(deepFreeze);
 };
 
+// workerk don't need default server, deprecation warnings, or package manager support
+exports.loadConfigWorker = (persistent, opts) => {
```
goshakkk: It's like 300ms faster on our app than `loadConfig` due to dropped unneeded steps

lib/commands.js
```diff
@@ -30,6 +30,7 @@ program.command('build [path]')
   .option('-e, --env [setting]', 'specify a set of override settings to apply')
   .option('-p, --production', 'same as `--env production`')
   .option('-d, --debug [pattern]', 'print verbose debug output to stdout')
+  .option('-J, --parallel [num]', 'parallelize the build')
```
goshakkk: Only in `build` for now

lib/fs_utils/generate.js
```diff
@@ -122,12 +122,21 @@ const runOptimizer = (optimizer, params) => {
...
+  const _optimize = () => {
+    return helpers.promisifyPlugin(1, optimizer.optimize)
+      .apply(optimizer, optimizerArgs);
+  };
+
+  const scheduleOptimize = () => {
+    return global.wrkr.schedule('optimize', {optimizer: optimizer.constructor.brunchPluginName, params: {data,path}})
+  };
+
+  return (global.wrkr ? scheduleOptimize : _optimize)()
```
goshakkk: Choose between original implementation and scheduling a worker job, depending on whether workers are enabled

lib/fs_utils/pipeline.js
```diff
@@ -164,20 +176,20 @@ exports.pipeline = (path, linters, compilers, fileList) => {
...
     };
 
     return cachedRead(path).then(path => {
       return Promise.resolve(path);
     }, (err) => {
       return Promise.reject(prepareError('Read', err));
-    }).then(callLinters)
-      .then(callCompilers)
-      .then(deppack.exploreDeps(fileList));
+    }).then(global.wrkr ? scheduleCompile : _compileAndLint).then(deppack.exploreDeps(fileList));
```
goshakkk: Same

lib/fs_utils/write.js
```diff
@@ -131,17 +131,13 @@ const write = (fileList, config, joinConfig, optimizers, startTime) => {
...
-      return Promise.reject(err);
-    });
+  const gen = file => {
+    return generate(file.path, file.sourceFiles, config, optimizers);
   };
 
   return Promise.resolve(changed)
     .then(changedFiles => {
-      return changedFiles.reduce(reduceGenerator, Promise.resolve());
+      return Promise.all(changedFiles.map(gen));
```
goshakkk: We need this so that optimizations are actually run in parallel, not in sequence... otherwise, there would be no gain in parallelizing these

lib/plugins.js
```diff
@@ -230,7 +230,7 @@ exports.init = (config, onCompile) => {
   const teardownBrunch = () => {
     teardowners.forEach(plugin => plugin.teardown());
   };
-  if (worker.isWorker) {
+  if (false && worker.isWorker) {
```
goshakkk: Previous worker implementation disabled...

lib/watch.js
```diff
@@ -270,7 +275,8 @@ class BrunchWatcher {
       // exit process with correct exit code.
       if (!config.persistent) {
         watcher.close();
-        worker.close();
+        setTimeout(() => notAFunction(), 10); // for some reason, the process won't exit until crashed even if each child process is killed...
```
goshakkk: :imp:

lib/worker-manager.js
```diff
@@ -0,0 +1,89 @@
+'use strict';
+const child_process = require('child_process');
+const events = require('events');
+const debug = require('debug')('brunch:workers');
+
+const workerFile = `${__dirname}/workeryo.js`;
```
goshakkk: needs a rename

lib/worker-manager.js
```diff
@@ -0,0 +1,89 @@
...
+  constructor(persistent, options, config) {
+    this.jobs = [];
+    this.workers = [];
+    this.pending = {};
+    this.free = [];
+    this.events = new events;
+    this.options = options;
+    this.config = config;
+
+    const num = typeof options.parallel === 'string' ? parseInt(options.parallel) : require('os').cpus().length;
```
goshakkk: Using `os.cpus.length` is not always reliable. On some newer macbooks, OS X reports double the original number of cores because "hyper threading", however, I've found that 4 workers are actually way slower than 2 on my macbook. There doesn't seem to be a way to get the real number of cores, so it's better to always pass `-J num`

lib/worker-manager.js
```diff
@@ -0,0 +1,89 @@
...
+    debug(`Spinning ${num} workers`);
+    for (let i = 0; i < num; i++) {
+      this.fork();
+    }
+
+    setInterval(() => this.sendMessage(), 1);
```
goshakkk: Try to send a first message from the queue to any available worker

lib/worker-manager.js
```diff
@@ -0,0 +1,89 @@
...
+    const options = Object.assign({}, this.options, {parent: null});
+    // pass parsed options to not make each worker parse the options
+    const workerEnv = {BRUNCH_OPTIONS: JSON.stringify(options)};
+    const env = Object.assign({}, process.env, workerEnv);
+    const wrk = child_process.fork(workerFile, {env});
+    const events = this.events;
+    const free = this.free;
+    let idx;
+    wrk.on('message', function(msg) {
+      if (msg === 'ready') {
```
goshakkk: Means worker already loaded the config and is ready to process jobs

lib/worker-manager.js
```diff
@@ -0,0 +1,89 @@
...
+        list.push(this);
+        idx = list.indexOf(this);
+        pending[idx] = false;
+      } else {
+        const id = pending[idx];
+        pending[idx] = false;
+        events.emit(id, msg);
```
goshakkk: Process a reply from worker, mark the worker as non-busy

lib/commands.js
paulmillr: Let's remove a CLI flag. That should be specified in configs. Let's keep CLI simple.

goshakkk: > It is better than storing that in the config, because number of workers that yield the speed increase depends on each machine, it's number of cores, so it doesn't belong to the config.

So why do you think it belongs in config? It's **machine dependent**, not so much **project dependent**, after all. It should be in the config. Some machines would be better off with no worker, some with 2, some with 4...

*(Make does it this way, [too](https://www.gnu.org/software/make/manual/html_node/Parallel.html))*

goshakkk:
@paulmillr take a look

es128:
Have you perf tested this? I eventually realized after my initial experiment with this was that the primary fatal flaw in what I had put together was using `process.send()` on the compiled results which you appear to still be doing. `process.send()` is a blocking sync operation (at least it was at the time) and passing around objects of that size made the performance worse the larger the project (where you'd expect to see the most perf gain from parallelism). I originally thought it was just the initial penalty from having to re-process the config and stuff like that, but that's not the most substantial problem.

Play with it and see if you can find circumstances where using this boosts speed. Better with larger projects than just plain skeletons.

The solution to the problem I've just described that I think merits looking into is to use streaming stdio interfaces to communicate between the processes and avoid `process.send()` for anything more than trivial signals (or perhaps avoid it altogether).

goshakkk:
@es128 as I've stated in the PR description,

> pipes would not provide much benefit compared to [cp/process].send **(which is also async in node 5)**
it's not like any tangible amount of time is spent in IPC, maybe 200-400ms tops in the before-mentioned app (**2-4%**), so probably shouldn't be our priority when optimizing

Pipes with streamed IO however will introduce much more complexity for no feasible benefit, even if it made IPC take zero time (which it, admittedly, will not)

---

> Have you perf tested this?

If you read into the PR description, the answer to that would be an apparent yes.

> On one of our apps that takes about 15s to compile on 2.4.2, and 25s to compile production (uglified) build on 2.4.2.
> With 2 workers, on a dual-core macbook, the dev build is 12s and production one is 20s (so, **~20% increase**).

--- 

>  I originally thought it was just the initial penalty from having to re-process the config and stuff like that, but that's not the most substantial problem.

For the reference, on the said app, the config takes about a second to parse and such, the optimized config load function for workers made it 200-300ms less. As explained in the PR, there doesn't seem to be a way to avoid that completely, unfortunately.

es128:
I apologize. You're right, I didn't read through everything you had written. Just skimmed very quickly and jumped straight to the code to look for `process.send`. Didn't remember I had written #880.

Good work, looks like you've done a thorough job here.

> With 2 workers, on a dual-core macbook, the dev build is 12s and production one is 20s (so, ~20% increase).

Any experiments with greater numbers of workers?

goshakkk:
@es128 on a dual-core machine, more workers lead to a decline in speed boost. 3 is slower than 2 (13-13.5ms instead of 12s); 4 were on par with no workers at all (15s). In general, # of workers should equal to the number of cores, tops.

goshakkk:
I don't have a machine with more cores to see how that'd improve speed

es128:
Interesting, thanks. I've heard anecdotal references that for server applications using `cluster` some people have found via load testing that the ideal number of workers was slightly less than 2x cores. Probably highly dependent on various factors in system architecture and the application itself.

lib/fs_utils/generate.js
```diff
@@ -122,14 +123,23 @@ const runOptimizer = (optimizer, params) => {
...
+
+  return (mediator.wrkr ? scheduleOptimize : _optimize)()
     .then(optimized => {
       const isObj = toString.call(optimized) === '[object Object]';
       const optimizedCode = isObj ? optimized.data : optimized;
       const optimizedMap = isObj ? optimized.map : null;
+
       let newMap;
-      if (optimizedMap != null) {
+      if (optimizedMap != null && false) {
```
paulmillr: wut

goshakkk: A debugging leftover. Just removed it

PR #1223, opened 2016-03-08
Workers refactor idea

goshakkk:
My first shot at making using workers a little bit clearer.

The gist is formalizing which "pieces of computation" can be parallelized. Because only JSON can be sent between processes, the serialization/deserialization steps are now grouped in the "job description", instead of being scattered across switch in worker-job-processor and implementation in pipeline/generate.

A job is described by:

* `serialize`, which takes some `hash` and produces a hash to be sent as JSON (always executed in main process);
* `deserialize`, which takes the hash from the step above and should produce the **original** hash passed to serialize (always executed in worker process). To aid reconstructing original hash, `ctx` is passed (which contains all loaded plugins; we needed it to select appropriate plugins for a given file to compile)
* `work`: actually perform the work, can be called in both master and worker.

(Mediator is still used but it's not scattered across several files anymore. We can get rid of it, but the focus of the PR is on making workers easier to comprehend.)

Then, wherever we used to have a `mediator.wrkr ? oneFn : anotherFn` switch, we can now simply call `processJob(JobDescription, {dataForTheJob})`, and that function will itself figure out whether it needs to schedule work in a worker, or call job.work right away (if workers are disabled).

This is mostly an implementation of an idea. Any feedback and suggestions welcome.

cc @paulmillr @es128 

- [x] eslint
- [x] squash
- [x] no mediator


paulmillr:
Norm.

goshakkk:
Got rid of the mediator for workers

PR #1240, opened 2016-03-17
Compilers JS exports

goshakkk:
Allows non-JS compilers to additionally expose a JS version of a file (use case: CSS modules, the compiler needs to both pre-process the source file + obfuscate class name, as well as create a JS module with class name mappings)

To achieve that, plugins that used to return `{data, map}` can now return `{data, map, exports}` where `exports` is a string containing javascript code:

```javascript
class MyCompiler {
  compile(params) {
    const path = params.path;
    const data = params.data;

    const compiled = magic(data);
    const exports = "module.exports = 42";

    return Promise.resolve({ data: compiled, exports });
  }
}
```

Internally it is achieved by allowing a single file have multiple *targets* — at max, one per file type (javascript, stylesheet, template), which is enough to get the example above working. It's not public API so should we need more flexible architecture some time later, we can also have multiple outputs of the same file type or something.

A less flexible option would be to hardcode store `jsExports` on source file, which also requires more code to make this "special case" work. (See fa762b8 for hardcoded jsExports vs targets)

To do:

- [x] allow plugins to emit JS modules for non-JS files
- [x] introduce the notion of targets internally
- [x] check if existing APIs were not affected:
  - [x] `onCompile` callback and the `sourceFiles` it receives (without JS-exported ones; create and document a new attribute for these in the callback))
  - [x] make sure resolving file's dependencies still works
- [x] document the `onCompile` callback changes
- [x] document the plugin API changes
- [x] make sure changing a style file also causes its JS exports to update
- [x] update brunch-maintained style plugins to have an option to produce CSS modules + JS exports with mappings:
  - [x] css-brunch https://github.com/brunch/css-brunch/pull/8
  - [x] stylus-brunch https://github.com/brunch/stylus-brunch/pull/43
  - [x] less-brunch https://github.com/brunch/less-brunch/pull/27
  - [x] sass-brunch https://github.com/brunch/sass-brunch/pull/104

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -327,7 +327,16 @@ file watching library used in brunch.
 
 ## `onCompile`
 
-`Function`: Optional callback to be called every time brunch completes a compilation cycle. It is passed a `generatedFiles` array, as well as `changedAssets` array. Each member of `generatedFiles` array is an object with `path` (path of the compiled file) and `sourceFiles` (array of objects representing each source file), each member of `changedAssets` array is an object with `path` (original path of an asset) and `destinationPath` (path of an asset in the public directory).
+`Function`: Optional callback to be called every time brunch completes a compilation cycle. It is passed a `generatedFiles` array, as well as `changedAssets` array. Each member of `generatedFiles` array is an object with:
+
+* `path` — path of the compiled file
+* `sourceFiles` — array of objects representing each source file
+* `allSourceFiles` array of objects representing each source file — this one also includes files that don't belong to the original type (e.g. if a styles compiler adds a JS module, path would the the concated JS file, and one of the `allSourceFiles` will be a style file
+
+Each member of `changedAssets` array is an object with:
+
+* `path` — original path of an asset
+* `destinationPath` — path of an asset in the public directory
 
 Example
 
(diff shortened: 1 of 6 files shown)


colinbate:
This is great to see.

I wanted to import some external CSS using the import syntax in the JS file and found it worked, but the module wasn't added to the JS so it would error not found at run time. I was going to ask if there an official solution planned.

In the mean time I wrote colinbate/wrap-require-brunch to make the CSS require statement a no op in the JS file. It works but I like the flexibility of this as I was looking into CSS modules as the next step.

Let me know if you want help testing it out.

goshakkk:
@colinbate note that these changes will only help if your styles compiler produces JS exports. If it doesn't — then no module will be generated for these styles file so you'd still want to keep a no-op probably (also, compilers don't process vendor files)

We might add no-op requires for external stylesheets at some point, but the currently recommended way is with `config.npm.styles` (if they come from packages)

goshakkk:
@paulmillr now that the todos are completed and there are PRs for styles compilers adding css modules support, mind taking a look at this?

colinbate:
@goshakkk yes I was referring to the ability to have a plugin that output both stylesheets and JavaScript. Also the compilers do see the node_modules CSS files if required by your app.

paulmillr:
squash please

goshakkk:
Squashed

Issue #1247, opened 2016-03-21
Brunch does not compile dustjs templates


khamassn:
### Description

While using brunch along with dustjs-linkedin-brunch, dust template is not compiled.

### Environment

1. Brunch: 2.5.1
2. Node: 5.4.1
3. NPM: 3.3.12
4. Operating system: Xubuntu 15.04, Windows 7

### `package.json` contents
```
{
  "name": "title",
  "description": "description",
  "author": "",
  "version": "0.0.1",
  "license": "",
  "main": "./app/server.js",
  "repository": {
    "type": "git",
    "url": ""
  },
  "scripts": {
    "start": "node ./app/server.js",
    "debug": "NODE_ENV=development nodemon ./app/server.js --debug"
  },
  "dependencies": {
    "adaro": "^1.0.1",
    "auto-reload-brunch": ">= 1.0 < 1.9",
    "body-parser": "^1.14.2",
    "clean-css-brunch": ">= 1.0 < 1.9",
    "colors": "^1.1.2",
    "connect-flash": "^0.1.1",
    "cookie-session": "^1.2.0",
    "css-brunch": ">= 1.0 < 1.9",
    "dustjs-helpers": "^1.7.3",
    "dustjs-linkedin": "^2.7.2",
    "express": "^4.13.3",
    "file-stream-rotator": "0.0.6",
    "javascript-brunch": ">= 1.0 < 1.9",
    "lodash": "^3.10.1",
    "morgan": "^1.6.1",
    "passport": "^0.3.2",
    "passport-local": "^1.0.0",
    "q-io": "^1.13.2",
    "serve-favicon": "^2.3.0",
    "uglify-js-brunch": ">= 1.0 < 1.9"
  },
  "homepage": "",
  "devDependencies": {
    "brunch": "2.5.1",
    "dustjs-linkedin-brunch": "^1.7.3",
    "javascript-brunch": "^1.8.0",
    "sass-brunch": "^1.9.2"
  }
}
```

### brunch config contents
```
exports.config =
	paths:
		watched: ['app/assets', 'app/styles', 'app/scripts', 'app/views/precompiled', 'vendor']
	files:
		javascripts:
			joinTo:
				'app.js': /^app/
				'vendor.js': /^vendor/
		stylesheets:
			joinTo: 'app.css'
		templates:
			defaultExtension: 'dust'
			joinTo: 'app.js'
	server:
		port: 8138
		run: no
```

### Other useful files, when present (`bower.json` etc.)

[...]

PR #1254, opened 2016-03-22
Add plugins' includes as npm.statics

goshakkk:
This is to let deppack distinguish between modules and vendor files that
just happen to come from node_modules.

Reported in: https://github.com/brunch/with-react/issues/3

Partly related to: https://github.com/brunch/brunch/issues/1247 & https://github.com/brunch/deppack/pull/1

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -83,7 +83,7 @@ const applyOverrides = (config, options) => {
 
 const deepFreeze = object => {
   Object.keys(Object.freeze(object))
-    .map(key => object[key])
+    .map(key => object[key] && object[key] !== object.static)
     .filter(value => {
       return value && typeof value === 'object' && !Object.isFrozen(value);
     })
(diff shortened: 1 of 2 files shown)


goshakkk:
Also: https://github.com/brunch/deppack/pull/2

lib/config.js
```diff
@@ -83,7 +83,7 @@ const applyOverrides = (config, options) => {
 
 const deepFreeze = object => {
   Object.keys(Object.freeze(object))
-    .map(key => object[key])
+    .map(key => object[key] && object[key] !== object.static)
```
goshakkk: We don't want `config.npm.static` frozen just yet, because we need to be able to add `includes` from plugins to it after they are intialized

lib/watch.js
```diff
@@ -126,8 +126,11 @@ class BrunchWatcher {
     // Wish it worked like `watcher.add includes`.
     const rootPath = cfg.paths.root;
     this.plugins.includes.forEach(path => {
-      this.changeFileList(getRelativePath(rootPath, path), true);
+      const relPath = getRelativePath(rootPath, path);
+      cfg.npm.static.push(relPath);
+      this.changeFileList(relPath, true);
     });
+    Object.freeze(cfg.npm.static);
```
goshakkk: But we do freeze it once that is done

Issue #1252, opened 2016-03-22
Server path does not use public paths override after bump to 2.5.x


SuaYoo:
### Description

The default Brunch server does not use the `paths.public` override, and instead throws an `ENOENT: no such file or directory` error.  However, the files are correctly compiled into the directory specified in `overrides[setting].paths.public`.

### Environment

1. Brunch: 2.5.2
2. Node: 5.8.0
3. NPM: 3.7.3
4. Operating system: OS X 10.10.5

### `package.json` contents

```
{
  "devDependencies": {
    "brunch": "2.5.2",
    "handlebars-brunch": "2.0.1",
    "javascript-brunch": "2.0.0",
    "ng-annotate-brunch": "1.7.2",
    "sass-brunch": "2.0.0",
    "clean-css-brunch": "2.0.0",
    "uglify-js-brunch": "2.0.1"
  },
  "dependencies": {
    "dotenv": "2.0.0",
    "npid": "0.4.0"
  }
}
```

### brunch config contents

`preview` override in brunch-config.js:
```
module.exports = {
  config: {
    overrides: {
      preview: {
        files: {
          stylesheets: {
            joinTo: {
              'style.css': /^app\/sass\/preview/
            }
          }
        },

        paths: {
          public: 'build/preview'
        }
      }
    },

    npm: {
      enabled: false
    }
  }
}
```

### Additional information

This happens after bumping up from 2.4.2.

`brunch w -e preview -s -j 2` output:
```
22 Mar 18:45:27 - warn: config.files.templates.defaultPaths was removed
22 Mar 18:45:27 - info: application started on http://localhost:3333/
22 Mar 18:45:30 - info: compiling
22 Mar 18:45:34 - info: compiling.
22 Mar 18:45:38 - info: compiling..
22 Mar 18:45:42 - info: compiled 143 files into 4 files, copied 450 in 15.3 sec
Error: ENOENT: no such file or directory, stat '/PATH/public/index.html'
  at Error (native)
```

localhost:3333 should be pointing to `'/PATH/build/preview/index.html'` instead of `'/PATH/public/index.html'`.

goshakkk:
https://github.com/brunch/brunch/pull/1255 should fix this. Thanks for your report!

PR #1255, opened 2016-03-23
Take into account public folder override when configuring server

goshakkk:
Closes GH-1252


MaximAbramchuck:
@goshakkk take a look at https://github.com/hellyeahllc/brunch/pull/1 before merge please

Issue #1253, opened 2016-03-22
Allow specifying an entry point js file and follow imports to include all relevant files


philer:
In #1248 I described a setup where I have a single js file (`js/index.js`) that serves as an entry point to my app. That file contains mostly ES6 `import` statements that lead to other files. It would be great if I could tell brunch to compile this file and everything imported by it. Other build tools work that way (e.g. in webpack's configuration you specify an `entry` key), so this could also help projects migrate their code to brunch.

In the case of ES6's static `import`/`export` this should be quite simple. Things may become more complicated if the loader API is used which provides dynamic imports (`System.import(…)`). I'm not sure about CommonJS or AMD, it would be cool if these could be supported as well.

Another great feature for static imports would be tree-shaking. The [rollup.js](http://rollupjs.org/) project is built on this feature. Basically through static code analysis they not only determine which modules are used but also which individual exports are used. Any parts of an imported module that are not used will be excluded from the bundle (functions, variables etc.). Webpack2 [also offers a similar feature](http://www.2ality.com/2015/12/webpack-tree-shaking.html), however they use dead-code-removal which is different from tree-shaking as explained in [this post](https://medium.com/@Rich_Harris/tree-shaking-versus-dead-code-elimination-d3765df85c80#.l1mrlksjn) by the creator of Rollup.

goshakkk:
A feature like rollup could indeed be useful, but I have some reservations as to how well it can play with brunch's deppack. I probably can look further into it to analyze if it's doable and will work with existing infrastructure... 

---

@paulmillr do you think Rollup is something worth pursuing?

Also re entry point/imports: following imports inside app should be more or less easy to implement in Brunch given that it already does that for node_modules. 

The feature would entail:

* changes to Brunch, behind a config flag, to start with only one file from `app` rather than all of them
* probably should not require changes in deppack

I think the shortcomings would be, one entry point can't serve both app bundle and test bundle... 

@paulmillr do you think entry point is the concept we want to have in brunch in general and behind an opt-in flag in particular?

philer:
If this works out it might be worth setting it up so multiple entry points can be specified. Overlapping dependencies should only be included once.

es128:
[...]

PR #1258, opened 2016-03-24
Entry points

goshakkk:
Also https://github.com/brunch/deppack/pull/3

#1253 

This PR adds a concept of an entry point that is an alternative to `joinTo`. joinTo is as simple as it gets and flexible, but often enough, what you'd want is simply a bundle (or a few) that start out in some files and include all resolved dependencies of these files.

#### Some assumptions about and constraints of, entry points

- only the file and its explicit dependencies will be included (`require('something')`)
  if you have DSL style functions that require and do something, there's no way for deppack to know if it's a dep (e.g. `match('/', 'app/Home')` — is not statically analyzable)

  other non-module vendor files also won't be included in the bundle

- an entry point can be output to a single file or split between several files (resembling `joinTo`)
  ```javascript
  javascripts: {
    'app/initialize.coffee': 'javascripts/bundle.js'
  }
  // or
  javascripts: {
    'app/initialize.coffee': {
      'javascripts/app.js': /^app/,
      'javascripts/vendor.js': /^(?!app)/
    }
  }
  ```
- two entry points **can't** both write to a single file

  ```javascript
  javascripts: {
    'app/initialize.coffee': 'javascripts/bundle.js',
    // INVALID
    'app/bookmarklet.coffee': 'javascripts/bundle.js'
  }
  ```

- all `config.npm.globals` will be included in **EVERY** entry point and `joinTo` file

- entry points can only be used with javascript files; stylesheets are obviously out; templates can still `joinTo` to these files

#### To Do

- [x] basic working version
- [x] warn when a joinTo and entryPoints have the same target file name
- [x] don't allow a shared target file name between `templates` and `javascripts`
- [x] make sure entryPoints can only be used with javascripts
- [x] make sure entryPoints did not break joinTo/stylesheets
- [x] overrides
- [x] config slashes
- [x] come up with & formalize the implications of entry points with re deppack and define constraints
- [x] ensure uniqueness of entry point output files
- [x] templates
- [x] rebase on plugins' js-exports (now in master)
- [x] docs
- [x] generation log shouldn't be broken
- [x] refactor


paulmillr:
can you describe the concept in 20 lines of code / demo app?

goshakkk:
@paulmillr 

```javascript
files: {
  javascripts: {
    entryPoints: {
      'javascripts/app.js': 'app/init.js'
    }
  }
}
```

instead of 

```javascript
files: {
  javascripts: {
    joinTo: 'javascripts/app.js'
  }
}
```

which will include into the final bundle only the things you actually require.

Florian-R:
@goshakkk Can this have in some way the current behaviour of Brunch (splitting app and deps in two bundle) or you always end up with one monolithic big file?

goshakkk:
@Florian-R As it stands, this will currently create one big bundle, which is probably good enough for the first iteration. Do you have any ideas as to what a config for that would look like?

cc @es128

goshakkk:
I think one way to implement that might be switching places of target and entry point

```javascripts
files: {
  javascripts: {
    entryPoints: {
      'app/init.js': 'javascripts/app.js'
    }
  }
}
```

This will allow then to also pass an jointTo-like object instead of a string

```javascripts
files: {
  javascripts: {
    entryPoints: {
      'app/init.js': {
        'javascripts/app.js': /^app/,
        'javascripts/vendor.js': /^(?!app)/
    }
  }
}
```

goshakkk:
Which means that an entryPoint becomes a joinTo with an additional pre-filtering step. An interesting implication of that is that the regular joinTo can also be implemented like a special case of an entry point — the one that just uses all files available, instead of checking whether they're used

goshakkk:
@paulmillr @es128 mind taking a look? I have outlined the limitations and added the docs.

I will squash once I get and address the feedback 

Related: https://github.com/brunch/deppack/pull/3

lib/config.js
```diff
@@ -150,79 +169,130 @@ const replaceConfigSlashes = exports.replaceConfigSlashes = config => {
...
+};
+
+const normalizePluginHelpers = (items, subCfg) => {
+  const vendorRe = /vendor/i;
+  if (!subCfg) return;
+  if (!items) {
+    items = (() => {
+      const destFiles = Object.keys(subCfg);
+      const joinMatch = destFiles.filter(file => {
+        return subCfg[file]('vendor/.');
```
paulmillr: shouldn't this be more generic

lib/config.js
```diff
@@ -150,79 +169,130 @@ const replaceConfigSlashes = exports.replaceConfigSlashes = config => {
...
+      });
+      if (joinMatch.length > 0) {
+        return [joinMatch[0]];
+      }
+      const nameMatch = destFiles.filter(file => vendorRe.test(file));
+      if (nameMatch.length > 0) return [nameMatch[0]];
+      return [destFiles.shift()];
```
paulmillr: Let's drop `shift()`, this is mutable operation that has always been too weird for me.

es128:
In addition to processing the concat list, should use of `entryPoints` cause implicit `autoRequire` behavior that executes the entry module as soon as it's loaded?

paulmillr:
I think so, yep.

goshakkk:
The problem would be that `autoRequire` is global — meaning all js files will get all the requires. For example, with multiple entryPoints, that would result in both output files requiring both entry points; or with a joinTo and one entryPoint — in joinTo file getting a require for entry point — clearly not something we'd want.

In addition, that could also backfire if an entryPoint is code-split between several files. AutoRequire in that case would try to require as soon as the first file is loaded, which most of the time will result in a require error — because the other file with modules wasn't loaded yet (vendor or app, for example)

lib/config.js
goshakkk: it's just the existing code moved into a separate function. to not introduce any breaking changes in this (probably not oft used) part, i think it very well makes sense to keep it as is

lib/config.js
goshakkk: replaced with `destFiles[0]`

es128:
@goshakkk so then about in the case that the separate bundles are not intended to be used on the same page? What options are we leaving for users regarding how to get their code executing aside from writing initializing code into the template?

How can we provide options that allow for addressing these different use cases elegantly?

goshakkk:
@es128 what we could do in this case is have an entryPoint-aware code in `generate` that will add only the necessary require statement. Then we should enforce some assumptions as to which (if multiple output files) will contain the require statement so that the users know to include it last (maybe last defined output file for that entry point?)

Do you think this behavior should be default and how could it be enabled/disabled manually (from the config perspective)?

es128:
As you've pointed out, users may have several different approaches so it's a tough problem. Might need to have a separate `autoRequire` config key within `entryPoints` definitions in order to deal with the different modes that might be desired. I don't have a clear answer in mind on how to design it.

goshakkk:
It's going to snow-ball, isn't it? Maybe we should first release this feature as is, without a way to automatically require anything, and then listen to the user feedback and their various approaches, and come up with the most elegant way.

As it stands, we probably don't have enough data on how to implement it best, thus it might make sense to withdraw from implementing this one for now,.

The scope of the current PR seems to be well-defined and documented, and we can always tweak/add new features later.

es128:
Now that we've talked it through a bit, I fully agree :point_up_2:

es128:
I guess the one note is that we could try to understand webpack's approach to this a bit better and see if there are any insights to be learned.

Not that I'm volunteering to make that assessment.

goshakkk:
Webpack always starts with an entry point and always produces a single bundle (afaik) so they probably aren't having this issue

es128:
My understanding is that they have multiple entry point support, but I don't know much about how it's actually handled.

philer:
Webpack has [code splitting](https://webpack.github.io/docs/code-splitting.html). I dunno much about it either but maybe that link helps. ^^

es128:
@philer thanks, but I'm pretty sure that's a whole different topic for the most part. Actually, looking at the top of https://webpack.github.io/docs/multiple-entry-points.html - what becomes clearer is that this problem doesn't exist for them because entry point bundles are not designed for allowing the use of multiple on the same page. 

Which actually supports the original idea that entry point bundles should always automatically load the entry module (maybe we can just provide a simple boolean for disabling this if needed), and any of the use cases @goshakkk described that would be in conflict are actually just an inappropriate use of entry points. Users that need more fine-grained control should go back to `joinTo` and control the loading mechanisms themselves. Or we eventually move into our own code splitting implementation to allow for that sort of thing - but the important point is that is its own separate concept.

philer:
I'm not quite sure how this is supposed to work now. Here's what I'd be expecting:

```coffee
  files:
    javascript:
      joinTo:
        'out/all.js': /^app\//
      entryPoints:
        'out/single.js': 'app/singleEntry.js'
        'out/multi.js':  ['app/multi1.js', 'app/multi2.js', 'singleEntry.js']
```
So after building you would have three files in `out/`: `all.js` contains everything from the `app/` directory, `single.js` contains only `app/singlEntry.js` and all of its dependencies and `multi.js` contains the three files specified in the array (anymatch set) and all of their combined dependencies.
That way `single.js` is a subset of `multi.js` which is a subset of `all.js`.

Also I'm not quite sure but does this approach also cover ES6 import/export syntax? Cause that's kind where it's at these days. x)

Don't hesitate to tell me to shut up if this isn't the right place.

es128:
@philer Yes, that demonstrates a few different ways to use it, but I think most real-world cases would use only one of those modes at a time - or at least they wouldn't be configured to have so much overlap with each other.

nblackburn:
This would make writing isolated tests MUCH easier and potentially allow me to use Wallaby in Brunch projects which is super exciting to me.

zedryas:
Hello,

quick question, i like to use stylus, and stylus resolve it's dependencies just like javasctipt files with imports.

How come this entryPoints key is incompatible for stylesheets ?
```
 warn: entryPoints can only be used with 'javascripts', not 'stylesheets'
```

That option will be great for stylus for instance

PR #1266, opened 2016-03-30
Add ability to specify 'hooks' object in brunch-config

MaximAbramchuck:


es128:
:+1: makes sense

what plans are there for additional hooks beyond `onCompile`?

paulmillr:
We just need to understand the common use cases. I'm not sure since i've never used any hooks...

PR #1274, opened 2016-04-02
Fix auto-npm install. Make sure Brunch waits for it before proceeding

goshakkk:


lib/watch.js
```diff
@@ -60,7 +60,7 @@ const checkProjectDependencies = cfg => {
       const pkgs = out.error.filter(x => x.indexOf(':') !== -1).map(x => x.split(':')[0]);
       // filter out symlinked packages
       const isNotSymlink = pkg => isSymlink(pkgPath(pkg)).then(x => !x);
-      helpers.asyncFilter(pkgs, isNotSymlink).then(unmetPkgs => {
+      return helpers.asyncFilter(pkgs, isNotSymlink).then(unmetPkgs => {
```
goshakkk: Absence of this forgotten require caused Brunch to load and start building before auto-install had a chance to even finish. In case of absent `node_modules` directory, this meant brunch simply will run, happily compile nothing, and then if run in watcher, install node_modules after some amount of time... definitely not desired behavior

Issue #1284, opened 2016-04-06
Regression: Requiring JSON is broken in 2.5.0


abogaart:
### Description

Issue #1236 has been [reintroduced in v2.6.3](https://github.com/brunch/brunch/commit/df88f11f4985fa3f94676e825b3e348fcebb6bfd#commitcomment-16998229)

PR #1286, opened 2016-04-07
Call isNpmJSON directly from the deppack object to prevent undefined …

marcioj:
…error

deppack assigns to [itself some methods](https://github.com/brunch/deppack/blob/master/index.js) when the configuration is loaded for the first time. 
Using the reference for `isNpmJSON` in the file top level returns undefined, since `deppack.loadInit` was not called. I changed the code to call the method directly from the deppack object to prevent this race condition issue.


goshakkk:
👍 

Also closes https://github.com/brunch/brunch/issues/1284

Issue #1289, opened 2016-04-11
Should `config.conventions.assets` apply to npm stuff? / Difficult to use the JS of  the bootstrap-sass npm package


lydell:
### Description

It’s not possible to `require('bootstrap-sass')` (`bootstrap-sass` is an npm package) using the default Brunch config.

### Environment

1. Brunch: 2.6.5 (both globally and locally)
2. Node: 5.9.1
3. NPM: 3.7.3
4. Operating system: Ubuntu 15.10

### `package.json` contents

```json
{
  "private": true,
  "devDependencies": {
    "brunch": "^2.6.5",
    "javascript-brunch": "^2.0.0"
  },
  "dependencies": {
    "bootstrap-sass": "^3.3.6"
  }
}
```

### brunch config contents

```js
module.exports = {
  files: {
    javascripts: { joinTo: "app.js" }
  }
}
```

### Steps to reproduce

Setup:

```
$ mkdir brunch-test
$ cd brunch-test
$ echo '{"private": true}' >package.json
$ npm i -D brunch javascript-brunch
$ npm i -S bootstrap-sass
$ echo 'module.exports = {files: {javascripts: {joinTo: "app.js"}}}' >brunch-config.js
$ mkdir app
$ echo 'require("bootstrap-sass")' >app/app.js
```

Notice how `bootstrap.js` is copied as an asset instead of being wrapped as a CommonJS module in `public/app.js`:

```
$ brunch b
11 Apr 12:10:54 - info: compiled app.js into app.js, copied bootstrap.js in 400ms
$ grep bootstrap public/app.js
require("bootstrap-sass")
```

bootstrap-sass’s `package.json` contains `"main": "assets/javascripts/bootstrap.js"`. Disabling Brunch’s “assets” convention (or renaming the `assets/` directory in `node_modules/bootstrap-sass/`) gives the result I expect:

```
$ echo 'module.exports = {files: {javascripts: {joinTo: "app.js"}}, conventions: {assets: []}}' >brunch-config.js
$ brunch b
11 Apr 12:11:51 - info: compiled 2 files into app.js in 480ms
$ grep bootstrap public/app.js
require.register("bootstrap-sass/assets/javascripts/bootstrap.js", function(exports, require, module) {
  require = __makeRelativeRequire(require, {}, "bootstrap-sass");
 * Bootstrap v3.3.6 (http://getbootstrap.com)
 * http://getbootstrap.com/javascript/#transitions
[...]

PR #1291, opened 2016-04-12
Don't categorize node_modules as 'assets' even if they match the regexp

goshakkk:
Closes GH-1289

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -9,7 +9,8 @@ const coffee = require('coffee-script');
 const debug = require('debug')('brunch:config');
 const deepAssign = require('deep-assign');
 
-const loadInit = require('deppack').loadInit;
+const deppack = require('deppack'); // isNpm
+const loadInit = deppack.loadInit;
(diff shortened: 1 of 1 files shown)


lib/config.js
```diff
@@ -9,7 +9,8 @@ const coffee = require('coffee-script');
 const debug = require('debug')('brunch:config');
 const deepAssign = require('deep-assign');
 
-const loadInit = require('deppack').loadInit;
+const deppack = require('deppack'); // isNpm
```
goshakkk: @paulmillr the `isNpm` call **has** to happen on the deppack object, don't change it to `const isNpm = require('deppack').isNpm;`

paulmillr: Why? I am not a big fan of implicit "this". Can we change it?

goshakkk: No, it's not about `this`. It's about the fact that most of deppack functions only become exposed after the call to `loadInit`: see https://github.com/brunch/deppack/blob/master/index.js

PR #1296, opened 2016-04-15
Added support to pass custom command line args with `--`.

bigp:
This isn't a full ideal solution, but at least it allows the parameters past "--" to get leaked down to your custom script's `process.argv` array (includes the ones beforehand however, too).
Ideally these would be stored in some global / accessible property so that they can be easily obtained in a cleaned up array.

Example:

Call:

    brunch build -p -- "hello world"

Could then give you, ideally:

   exports.customArgs[0] == "hello world"

But for now you have to read the `process.argv` array manually and do what you want with it.

### DIFF
--- a/lib/commands.js
+++ b/lib/commands.js
@@ -31,6 +31,7 @@ program.command('build [path]')
   .option('-p, --production', 'same as `--env production`')
   .option('-d, --debug [pattern]', 'print verbose debug output to stdout')
   .option('-j, --jobs [num]', 'parallelize the build')
+  .option('--', 'separates and passes the remaining parameters to your brunch-config script')
   .action(commands.build);
 
 program.command('watch [path]')
@@ -42,6 +43,7 @@ program.command('watch [path]')
   .option('-d, --debug [pattern]', 'print verbose debug output to stdout')
   .option('-j, --jobs [num]', 'parallelize the build')
   .option('--stdin', 'listen to stdin and exit when stdin closes')
+  .option('--', 'separates and passes the remaining parameters to your brunch-config script')
   .action(commands.watch);
 
 const checkForRemovedOptions = (args, command) => {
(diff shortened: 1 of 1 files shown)


bigp:
Apologies in advance if I'm not making pull-requests correctly (my 1st to open-source projects to be honest!)

If I'm supposed to bump up the version number of something (package.json or whatever), let me know.

paulmillr:
What are you trying to achieve with this? Could you give us any examples?

bigp:
Best way I can describe it (since my project example is way too cluttered to send at the moment), is that I have a brunch project that uses a `brunch-config.js` file requiring custom modules (although I wouldn't call my scripts modules at this point, still early in development!).

I call it a `multiout` setup, where I'm publishing out several \*.html, \*js, \*.css and \*.png (texture atlases) for my HTML5 ***ad development***. And my client recently demanded a whole bunch of them, in different categories, and languages. Fun stuff.

I'm aware there are overwriteable `--env` / `-e` command-line options, but since I want to keep the **production** vs **debug** (or however you call the `brunch watch --server` default environment), but I prefer to keep those parameters apart to strictly tell my app when to publish all the Ads for "devevelopment" purpose, and when to optimize it for "production" (strip out any console.log / debug checks etc.).

How I use the extra custom parameters is to focus brunch on an entire different sub-directory (essentially a specific Ad category, different campaigns type of thing) to automatically populate it's watch-folders, `js` and `css` **joinTo** file patterns.

That way, instead of dumping all the ads directly under `app/`, I can organize them as `app/campaignA/` and `app/campaignB/` and selectively compile only a smaller batch of Ads that I need, by typing something like:

    brunch watch --server -- campaignA

Now, how my config file uses the "campaignA" parameter is a very complex setup I could post in an entire article of its own. But I think all in all, allowing devs to use their own custom parameters within a brunch command-line call is a bonus.

Hopefully this makes... some sort of sense.

paulmillr:
```
brunch watch —server —env campaignA
brunch build —production —env campaignA
brunch build —env production campaignA
```

all of those should work.

bigp:
Ah! Well then... I take back my pull-request! Sorry about the huge unecessary explanation!

Going to close this now.

squishleheimer:
Can we get a concrete example of how to use/access the CLI --env variable?

paulmillr:
Use where?

PR #1298, opened 2016-04-16
Hot Module Replacement API

goshakkk:
To do: 

- [x] refactor
- [x] docs
- [x] actually reload the page if updates can't be applied (how to make sure the reason for the reload is kept? console is flushed on the reload)
- [x] update redux skeleton — https://github.com/brunch/with-redux/pull/2

After merged (in order):

- [x] release deppack
- [x] release commonjs-brunch
- [x] release auto-reload-brunch
- [x] bump deppack, commonjs; release brunch
- [ ] add `auto-reload-brunch` as `hmr-brunch`'s peerDep
- [x] release https://github.com/brunch/hmr-brunch
- [x] merge update for redux skeleton — https://github.com/brunch/with-redux/pull/2

What is intentionally missing from this PR:

- source maps for cases when `if (module.hot) { ... }` checks are cut — since production isn't going to need source maps anyway
- callback for self-accept `accept()` for when there is an error

Relevant PRs:

* [x] https://github.com/brunch/deppack/pull/19 (deppack changes needed to support)
* [x] https://github.com/brunch/commonjs-require-definition/pull/21 (HMR runtime code)
* [x] https://github.com/brunch/auto-reload-brunch/pull/70
- [x] https://github.com/brunch/with-redux/pull/2 (**only** after this is merged and all releases are cut)

---

http://github.com/brunch/hmr-brunch

### DIFF
--- a/lib/hmr.js
+++ b/lib/hmr.js
@@ -1,0 +1,40 @@
+'use strict';
+
+const sysPath = require('path');
+const logger = require('loggy');
+
+const isEnabled = config => config.hot && config.env.indexOf('production') === -1;
+
+const checkAutoReload = (cfg, plugins) => {
+  const hasAutoReload = plugins.compilers.find(compiler => compiler.brunchPluginName === 'auto-reload-brunch');
+  if (!hasAutoReload) {
+    throw new Error("Hot Module Reloading should only be used with 'auto-reload-brunch'");
+  } else if (!hasAutoReload.supportsHMR) {
+    throw new Error("Currently used version of 'auto-reload-brunch' does not support Hot Module Reloading");
+  }
+};
+
+const generate = (generateModuleFiles, nonModuleFiles, processor, deppack, definition, path, root) => {
+  const hmrFile = nonModuleFiles.find(x => x.path.indexOf(sysPath.join('hmr-brunch', 'runtime.js')) !== -1);
+  nonModuleFiles = nonModuleFiles.filter(f => f !== hmrFile);
+
+  if (hmrFile) {
+    processor(hmrFile);
+  } else {
+    logger.warn(`HMR runtime not found for ${path}. HMR only works if use compile your JS into a single file, could it be the case?`);
+  }
+
+  root.add(definition);
+
+  const depGraph = deppack.graph();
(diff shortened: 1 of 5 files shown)


lib/hmr.js
```diff
@@ -0,0 +1,40 @@
...
+  const depGraph = deppack.graph();
+  root.add(`require.hmr(${JSON.stringify(depGraph)}, function(require) {\n`);
+
+  generateModuleFiles();
+
+  root.add('});');
+  root.add('// hmr end');
+
+  nonModuleFiles.forEach(processor);
+};
```
goshakkk: I'm wondering if we should extract this piece of logic into the `hmr-brunch` plugin and introduce some *private* (for the time being) plugin API to plug into `generate` process. Thoughts? @es128 @paulmillr

lib/hmr.js
```diff
@@ -0,0 +1,40 @@
...
+const isEnabled = config => config.hot && config.env.indexOf('production') === -1;
+
+const checkAutoReload = (cfg, plugins) => {
+  const hasAutoReload = plugins.compilers.find(compiler => compiler.brunchPluginName === 'auto-reload-brunch');
+  if (!hasAutoReload) {
+    throw new Error("Hot Module Reloading should only be used with 'auto-reload-brunch'");
+  } else if (!hasAutoReload.supportsHMR) {
+    throw new Error("Currently used version of 'auto-reload-brunch' does not support Hot Module Reloading");
+  }
+};
```
goshakkk: This check can't be in the plugin initializer because it needs to be called only after all the plugins are loaded... again, it might make some sense to introduce some private API for a check like this, like `Plugin.prototype.checkInit`, or it might be an overkill...

goshakkk: Or maybe just specifying a `peerDependency` for hmr on newer version of auto-reload-brunch will do

goshakkk:
This PR is (and all its dependents are) ready for review @paulmillr @es128 

I'm mostly ok with the state of this. If anything, introducing some private plugin API for a thing or two (https://github.com/brunch/brunch/pull/1298#discussion_r60193084) could help better structure this functionality logically, but I didn't do that just yet, want to hear your feedback first.

PR #1299, opened 2016-04-18
Nitrous.io "Quickstart" Support

5car1z:
Hi Brunch!

Recently [Nitrous.io]( https://www.nitrous.io) have added a new feature to their platform named “Quickstarts”. This feature lets users create pre-configured development environments on the service at the click of a button. Nitrous are trying to set up these Quickstarts for users by requesting that projects like yourself host the Quickstart trigger buttons on your GitHub readme. 

This would mean that if a Nitrous user wants a dev environment all set up and ready to develop Brunch applications with - they can click the HTML button embedded on your readme, and then the Nitrous API creates one for them.

* https://www.nitrous.io/quickstarts/  - repositories so far we've got Quickstarts for. 

All that’s needed from yourselves is that you keep the three API configuration files (seen in these proposed changes) inside your repo's root, and also that you have the HTML trigger button embedded in the main `README.md` - for people to click (this is also included in these changes). 

Is this something you’d be up for helping us out with? It would really make using Brunch on the platform super easy. 

To be clear there’s not much you have to do on your end, except merge these changes into your master. We (likely me) would make any new pull requests to yourself in the future, to maintain or ensure that these three files are up to date. None of your actual project code on GitHub gets touched or is involved at any point. The files here simply tell the API what to do when the button is clicked by  a Nitrous user. 

Here’s some more links that might help explain:

* [Blog Post Explaining the Idea](https://community.nitrous.io/posts/nitrous-quickstarts)
* [Specifically How it Works](https://community.nitrous.io/docs/nitrous-quickstarts)

Any questions do ask. 

Thanks guys.

### DIFF
--- a/nitrous-post-create.sh
+++ b/nitrous-post-create.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+npm install -g brunch bower
+
+cd ~/code
+
+brunch new brunch-babel-es6 -s https://github.com/brunch/with-es6
+( cd ~/code/brunch-babel-es6 ; brunch build )
+
+brunch new brunch-exim-react -s https://github.com/hellyeahllc/brunch-with-exim
+( cd ~/code/brunch-exim-react ; brunch build )
+
+brunch new brunch-chaplin -s https://github.com/paulmillr/brunch-with-chaplin
+( cd ~/code/brunch-chaplin ; brunch build )
--- a/nitrous.readme.md
+++ b/nitrous.readme.md
@@ -1,0 +1,68 @@
+# Nitrous Brunch Quickstart
+
+This `nitrous.readme.md` file contains instructions on how to use the Nitrous [Quickstart](https://www.nitrous.io/quickstarts/) feature with this project.
+
+Quickstarts allow you to create an out of the box development environment on Nitrous in just a few minutes; without having to perform the expected setup or configuration beforehand.
+
+## How to Use
+
+All that's required is a [Nitrous.io account](https://www.nitrous.io) and that your are signed in to this account with your web browser when you click the Quickstart trigger button below.
+
+Click the button to begin the process once signed in:
+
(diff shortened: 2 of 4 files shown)


nitrous.readme.md
```diff
@@ -0,0 +1,68 @@
+# Nitrous Brunch Quickstart
```
paulmillr: let's put the file inside Docs, like that: `docs/nitrous.md`

nitrous-post-create.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
paulmillr: do we need the file?

5car1z: Ideally yeah as this is what installs Brunch onto the Nitrous dev environment that's being created for the user.

nitrous.readme.md
5car1z: I'll find out how moving these files affects the process internally, as at the moment we've been told to they need to be in the root of the Git project; should be okay though for the readme. Will get back to you on this.

goshakkk: Does this belong to the core repo? If anything, could we create a separate repo, perhaps something like `brunch-nitrous` that will contain this doc and the scripts? In which case, they can be at the root, not in a subdir

As not to pollute the main repo... What do you think?

colinbate:
I've started adding the Nitrous Quickstarts to a couple of my apps (which use Brunch) and it feels like this would be a better fit to add to the skeletons rather than to the main brunch repo.

This example installs three arbitrary? skeletons and lets you start with them which seems a bit odd to me.

Perhaps a brunch-nitrous repo or doc file with instructions on adding Nitrous support to your skeleton with some Brunch specific information (like setting the hostname). To take it a step further the skeletons with support could even have a little button on the skeletons list page.

nitrous.readme.md
5car1z: That does sound better, makes a lot of sense. Quickstarts for specific Brunch skeleton repos can be listed and linked on there too - as suggested by [Colin](https://github.com/colinbate) below. If you create the new `brunch-nitrous` repo I'll submit another pull request to it with these proposed Quickstart files (and some amendments). 

That sound good? Or is there a better way?

PR #1306, opened 2016-04-20
Fix duplicate dependencies in package.json

VitaliyR:
Hello there!

I've found a bug with duplicating plugins in package.json. Imagine situation:

```javascript
"dependencies": {
  "my-cool-optimizer": "^0.0.1"
},
"devDependencies": {
  "my-cool-optimizer": "^2.0.0"
}
```

In this case, optimizer from my-cool-optimizer plugin would work two times. I think it is wrong, isn't it? And it can break _(and, actually, it break my build process)_ somebody build process if it is doing something with output (e.g. wrapping it).

**What do you think about it? Is it needed or I'm too captious?**

And, I think we need to have some _FAQ for developers_ - even starting from 'how to roll out dev env for contributing brunch' - because that loadBrunch && loadGlobalBrunch, which appeared recently, made my day for a while when I tried to debug brunch :)

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -73,9 +73,7 @@ const loadPackages = (rootPath) => {
 
   // Also need to test if `brunch-plugin` is in dep’s package.json.
   const loadDeps = (prop, isDev) => {
-    const deps = Object.keys(prop || {});
-
-    return deps.filter(dependency => {
+    return prop.filter(dependency => {
       return dependency !== brunchPluginPattern &&
         dependency.indexOf(brunchPluginPattern) !== -1;
     }).map(dependency => {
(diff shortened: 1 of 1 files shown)


goshakkk:
Yup, although you are the first to report a use case like this, I think for cases like this, we can print a warning and init the plugin only once.

Regarding a contributing document — yeah, I think that makes sense as well. I'll add that one to our backlog.

VitaliyR:
I added it to my `devDependencies` and on next day runned `npm i --save` it again - so it appeared in both places :) 

Should I add a warn at line 105 in my PR? In that `map` function body?

lib/plugins.js
```diff
@@ -98,12 +96,24 @@ const loadPackages = (rootPath) => {
...
+  const uniqueDeps = function(/*mainDeps, ...otherDeps*/) {
+    return slice.call(arguments, 0).map(deps => {
+      return Object.keys(deps || {});
+    }).map((dep, index, deps) => {
+      if (!index) { return dep; }
+      return dep.filter((dependency) => {
+        return deps[0].indexOf(dependency) === -1;
+      });
+    });
+  };
```
goshakkk: can this be simpler by using Set?

```
const concatedArguments = slice.call(arguments).reduce((acc, x) => acc.concat(x));
const set = new Set(concatedArguments);
return Array.from(set.values());
```

goshakkk:
Yeah, a warning would be 👍 

Something along the lines of "You have `my-cool-optimizer` declared in both `dependencies` and `devDependencies`"

lib/plugins.js
goshakkk: Hm, actually, to retain the `isDev` behavior, the current code kind of makes sense. A little less obvious for sure, though 😟

lib/plugins.js
```diff
@@ -98,12 +96,24 @@ const loadPackages = (rootPath) => {
...
 
-  const plugins = loadDeps(json.dependencies);
-  const devPlugins = loadDeps(json.devDependencies, true);
-  const optPlugins = loadDeps(json.optionalDependencies, true);
-  const allPlugins = plugins.concat(devPlugins, optPlugins);
+  const uniqueDeps = function(/*mainDeps, ...otherDeps*/) {
+    return slice.call(arguments, 0).map(deps => {
+      return Object.keys(deps || {});
+    }).map((dep, index, deps) => {
+      if (!index) { return dep; }
```
goshakkk: can this be `index === 0`?

lib/plugins.js
VitaliyR: Yes, it is. I tried to play differently, but w/o refactoring loadDeps I cant make it looks more nicer.

lib/plugins.js
VitaliyR: Okay, maybe it is more readable.

PR #1308, opened 2016-04-21
Fix sourcemaps when using custom wrapper

VitaliyR:
Hello there.

Using custom wrapper I found that there is code inconsistency. 

Please, look at:
line **34** in modules.js _(method **getModuleWrapper**)_
line **34** in source_file.js _(method **wrappedNode**)_

Everything will work fine if we will return a string, but custom wrapper defined by function returns object { data: string } which breaks source map for that file, because **identityNode** is not expecting the code will go from different file than passed _path_ parameter to it. 

So, we need:
* either to refactor identity node
* or allow custom wrappers to return prefix and suffix itself
* or just look at this PR

What do you think, guys?

### DIFF
--- a/lib/modules.js
+++ b/lib/modules.js
@@ -31,9 +31,7 @@ const getModuleWrapper = (type, nameCleaner, wrapper) => {
         data: data.replace(amdRe, match => '' + match + path + ', ')
       };
     } else if (type === 'function') {
-      return {
-        data: wrapper(moduleName, data)
-      };
+      return wrapper(moduleName, data);
     }
   };
 };


Florian-R:
@VitaliyR Apologies for the inconvenience. It seems I'm the one who break it during #1199

I have a project using custom wrapper, I'll try your branch and report if I see errors with your approach.

VitaliyR:
@Florian-R 👍  Let me know when you check it

Florian-R:
So I double check with your patch and with 2.6.6 and I've seen no real differences. Code and sources maps are OK in either case.

Can you expand on the failures you encounter with source maps? Are you on 2.6.6?

VitaliyR:
@Florian-R yes, 2.6.6.

This guy is breaking:

```javascript
wrapper: function (path, data) {
      var result = '';

      if (/utilities/.test(path)) {
        return data;
      } else {
        result += "(function (root, factory) {\nif (typeof define === 'function' && define.amd) {\ndefine([], factory);\n} else {\nroot." + path + " = factory();\n}\n}(window, function () {\n";
        result += data + '\n';
        result += 'return ' + path + '; \n';
        result += "}));";
      }

      return result;
    }
```

VitaliyR:
Source maps is generated, yes, but relation between minified and original sources are wrong.

Florian-R:
This is weird. Unless I miss something, Chrome reports me exactly the same file structure with both versions.

VitaliyR:
@Florian-R try to set a breakpoint (e.g. in index.js) and check it in some other file (e.g. ./src/mymodule.js) - it it setted wrong and debugging process is awful

goshakkk:
@VitaliyR Does this PR in any way fix that? Probably does, since if the wrapper returns a string, https://github.com/VitaliyR/brunch/blob/custom-wrapper-sm/lib/fs_utils/source_file.js#L36 will try and detect the prefix/suffix itself.

VitaliyR:
@goshakkk I can't get there. Because that function from config returns string which wrapped into `{ data: ''that returned string"` - and suffix & prefix are undefined.

goshakkk:
But with this PR, it will be a string, not a `{data}` object? In which case, the function I linked should infer the prefix/suffix itself.

I guess I'm more of asking whether this PR really fix the issue. From the look of it, it should.

VitaliyR:
@goshakkk yes, by this PR it would be a string. So wrapperNode func thinks it is not and object so it is trying to parse from it (line 36-37) prefix and suffix by slice. And correct source map is generated.

goshakkk:
@VitaliyR 👍 did you confirm it's working?

If so, @paulmillr it's safe to merge

VitaliyR:
@goshakkk yes, I'm using right now this version and it is working as expected to be.

goshakkk:
👌

PR #1311, opened 2016-04-23
Write the contributing document

goshakkk:
@VitaliyR in #1306 reminded us that we should ease the entry for these who wish to try and fix the bugs they are experiencing themselves, or add a new feature.

I've written a simple contributing document that gives bird's eye overview of the various components of Brunch, as well as includes the steps to use the cloned fork locally (since not everyone might have even heard of `npm` commands other that `install`)

What do you think? @paulmillr @es128 

@VitaliyR, do you think this would have helped you even a bit? Is there anything else you think we should make sure to include there?

I thinking about outlining the brunch build process (aka which function in which file is called to compile a file, which is called after it, etc)...

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,69 @@
+# Contributing to Brunch
+
+### Did you find a bug?
+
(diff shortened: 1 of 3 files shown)


VitaliyR:
@goshakkk Yes, it is good start 👍 
After merging this one I will add info about debugging too

Issue #1294, opened 2016-04-13
Let's make awesome validation for the config!


lydell:
Update/TLDR:

- [x] This issue was originally about validating brunch-config.js. That’s done (in a general way that ensures that future config is validated too).
- [x] Then, I suggested better error handling of missing files and valid but bad config. That’s done. (Let’s improve things more as we find them.)
- [ ] What’s left to do is plugin config validation.

---

The last couple of days I've been migrating a legacy project to brunch, and it's been a very positive experince!

I've used brunch a bit a few years ago, so it wasn't entirely new to me, but I still bumped into a few problems along the way. Most of the time, the problems arose from a mistake I had made in brunch-config.js. Here are a few examples.

---

I had just finished installing some brunch plugins and making a few edits to brunch-config.js. I then ran `brunch b` to test it out. (Yes, my primary testing consists of running `brunch b` and inspecting the output in `public/`.  I don't bother opening the browser until it looks ok.) This is the result:

```
$ brunch b
13 Apr 06:06:33 - error: Initialization error - TypeError: stylesheets.map is not a function
  at /home/lydell/myproject/node_modules/deppack/lib/load-globs-styles.js:21:26
  at Array.map (native)
  at Object.keys.map.each.then (/home/lydell/myproject/node_modules/deppack/lib/load-globs-styles.js:15:42)
  at /home/lydell/myproject/node_modules/deppack/lib/deppack.js:71:19
```

That's not very helpful! Initially, I had no clue at all what was wrong. I didn't know if it was my fault, some plugin's fault or a bug in brunch. I then noticed the word "stylesheets" in the error message, and thought of `config.files.stylesheets`, but that was a red herring. I then realized that if `.map` is not a function, that probably means that brunch expects an array in my config, but I gave it something else. I started commenting out my most recent changes in brunch-config.js until the problem went away. Turned out this was the problem:

```js
npm: {
  styles: {
    'font-awesome': 'css/font-awesome.css',
  },
},
```

... which is supposed to be (note the added `[` and `]`):

```js
npm: {
  styles: {
    'font-awesome': ['css/font-awesome.css'],
  },
},
```

---

Similarly to the above, I Initially thought that `npm.static` could be used like this:

```js
static: {
  'font-awesome': ['fonts/font-awesome.woff']
},
```

That gave me this:

```
$ brunch b
/home/lydell/myproject/node_modules/deppack/lib/deppack.js:34
[...]

PR #1312, opened 2016-04-23
Config validation

goshakkk:
This PR introduces config schema and types validation, as well as allows to centralize default values for various config keys via a simple descriptive DSL.

* errors and aborts brunch on incorrect types
* warns on deprecations
* warns on unrecognized `config.files` keys, for example, along with a list of possible values and suggestions

Sample error messages: 

screen shot 2016-04-21 at 2 42 02 pm

### A few (possibly breaking) changes

* `config.paths.watched` won't be coerced into an array if string. probably ok nevertheless since the type mismatch error will tell it should be an array

`config.modules = false` won't be possible... but then, it is not even documented, which means it shouldn't be relied up

`config.paths.*` will now always be relative to root. The current behavior is that if custom paths aren't specified, then they are joined with root, but custom paths *are not*. Seems a little inconsistent and all paths, default or custom, will now be relative to root.

for consistency, `preCompile` is also moved into `config.hooks.preCompile`, in line with `config.hooks.onCompile`

### To do

- [x] squash
- [ ] release skemata v0.1.0 https://github.com/brunch/skemata @paulmillr

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -13,6 +13,8 @@ const deppack = require('deppack'); // isNpm
 const loadInit = deppack.loadInit;
 const mdls = require('./modules');
 
+const skemata = require('skemata');
+
 const _helpers = require('./helpers');
 const isWindows = _helpers.isWindows;
 const replaceSlashes = _helpers.replaceSlashes;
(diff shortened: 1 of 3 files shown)


goshakkk:
Started out at https://github.com/brunch/brunch/issues/1294

@lydell ping, though you might want to have a look 😄

goshakkk:
(That test fails because `skemata` is not yet released on NPM.)

lydell:
Thanks for the ping! Using a schema was exactly what I had in mind. It looks really good! I'll try this out in a few days to see what the experience is like.

PR #1330, opened 2016-05-02
Failing test with config override regression

marcioj:
Since 2.7.2, looks like brunch is not correctly merging some values when using the env option. I created a failing test case to illustrate this issue.

In the second test: `overrides the config using the specified env` the expected value for `config.paths.watched` is `['app', 'test']` but instead it returns `['app', 'test', 'vendor']`. In 2.7.1 this test pass.

### DIFF
--- a/test/config.js
+++ b/test/config.js
@@ -1,0 +1,26 @@
+const path = require('path');
+const test = require('ava');
+const config = require('../lib/config');
+
+test('loads the config without overriding', t => {
+  const opts = {
+    config: path.relative(__dirname, './fixtures/config-with-overrides.js')
+  };
+
+  return config.loadConfig(false, opts, true).then((config) => {
+    t.is(config.paths.public, 'public');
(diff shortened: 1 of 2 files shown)


paulmillr:
travis check failed

es128:
@paulmillr It is intended to fail. To demonstrate the regression that needs to be fixed (i.e. the PR did not come with a fix, only the test)

goshakkk:
I've changed skemata to perform deep undefaulting which does fix this test. (https://github.com/brunch/skemata/commit/964582f9306fb4459815ad845602cf946273aaff)

@paulmillr can you release https://github.com/brunch/skemata and merge this PR?

marcioj:
I tested with the latest version and worked fine, thanks!

Would you guys accept some PR's improving the brunch test coverage?

goshakkk:
Sure. We also are planning on improving our testing strategy, primarily with acceptance tests

Issue #1336, opened 2016-05-04
AVA same method is deprecated


denar90:
### Description

In AVA `t.same` is deprecated and in console I can see smth like 
`DEPRECATION NOTICE: same() has been renamed to deepEqual() and will eventually be removed. See https://github.com/jamestalmage/ava-codemods to help upgrade your codebase automatically.`
`same()` has to be replaced by `deepEqual()` in [fs_utils.common.js](https://github.com/brunch/brunch/blob/571ea17e07bc8159bda621d7738f49e849c16944/test/fs_utils.common.js), [fs_utils.generate.js](https://github.com/brunch/brunch/blob/571ea17e07bc8159bda621d7738f49e849c16944/test/fs_utils.generate.js), [helpers.js](https://github.com/brunch/brunch/blob/571ea17e07bc8159bda621d7738f49e849c16944/test/helpers.js)

Should I create PR for this?

paulmillr:
yes

denar90:
Created PR #1339

PR #1339, opened 2016-05-04
replace same by deepEqual

denar90:
PR for issue #1336 
Running `npm test` I got `✖ 2 problems (1 error, 1 warning)` but looking at [build](https://travis-ci.org/brunch/brunch/jobs/127817268) I think this problem was present before, so I think, I didn't break  anything.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -631,6 +631,7 @@ const tryToLoad = (configPath, fallbackHandler) => {
   try {
     pkg = require(sysPath.resolve('.', 'package.json'));
   } catch (e) {
+    logger.error(e);
   }
 
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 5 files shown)


denar90:
Added some test fixes, but I feel that tomorrow I'll get more elegant solution.

lib/config.js
```diff
@@ -631,6 +631,7 @@ const tryToLoad = (configPath, fallbackHandler) => {
   try {
     pkg = require(sysPath.resolve('.', 'package.json'));
   } catch (e) {
+    logger.error(e);
```
goshakkk: We don't actually need this logged. We are loading package.json here simply to try and recover from brunch-config require errors

denar90: But tests are crashing with empty `catch`

goshakkk: Just adding a comment or something in there should have made eslint happy

like

```
} catch (e) {
  // no-op
}
```

denar90: go it

Issue #1281, opened 2016-04-05
Normalize relative paths early


colinbate:
Currently when running on Windows, the paths for files use backslashes for a part of the workflow, and eventually get normalized into forward slashes when written out as module names. Given the reliance on regular expressions for config and in various other places, I thought it might be a good idea to normalize the relative paths from the beginning in `watch.js` after they come in from `chokidar`.

That way anytime you find yourself writing `/^app\/something/` you don't need to wonder if you need to provide both slashes (or worse not adding both until someone raises an issue about Windows support).

paulmillr:
agree. that's a nice ux improvement. let's see what we can do here.

mcfarljw:
I was having some similar troubles with then with a plugin of mine. I ended up throwing together a quick hack to get it running, but certainly wouldn't be against standardizing it on a higher level.

https://github.com/mcfarljw/replace-brunch/commit/11032a96c2ae40c40c4613f990c8c9737e6e62e4

goshakkk:
Here is my stab on that: https://github.com/brunch/brunch/pull/1346

goshakkk:
@paulmillr we can close this.

PR #1346, opened 2016-05-06
Normalize paths to always use UNIX slashes internally

goshakkk:
Ref GH-1281

Allows to get rid of platform-specific path handling in config regexps,
et cetera, and also relieves plugins from having to care about that.

Filesystem access functions work just fine with the `/`s.

Paths consistency is acheieved by normalizing the paths that come from
the watcher and by providing a local drop-in replacement for the `path`
module which always returns paths with the `/`s.

---

Deppack PR: https://github.com/brunch/deppack/pull/30

@paulmillr to do:

- [x] merge https://github.com/brunch/deppack/pull/30
- [x] release deppack **0.5.0**
- [x] bump deppack to **0.5.0**
- [ ] merge this

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,6 +1,6 @@
 'use strict';
 const fs = require('fs');
-const sysPath = require('path');
+const sysPath = require('./path');
 const exec = require('child_process').exec;
 const logger = require('loggy');
 const readComponents = require('read-components');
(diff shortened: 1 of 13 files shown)


goshakkk:
I have tested this in a windows VM with a react skeleton and it seems to work just fine (w both this *and* deppack changes).

Probably isn't too great to duplicate this `path` shim in both brunch and deppack, but I feel like extracting that into a separate module could be an overkill (though not as much as having an `is-number` package 😄)

PR #1351, opened 2016-05-08
Create some build tests and fix an issue with destinationPath

marcioj:
While creating some additional tests for brunch I noticed a typo in the destinationPath property.
So I'm opening a PR here to validate if the way I'm testing looks good and also bring the destinationPath fix.

### DIFF
--- a/lib/fs_utils/asset.js
+++ b/lib/fs_utils/asset.js
@@ -45,7 +45,7 @@ class Asset {
   copy() {
     const path = this.path;
     if (isIgnored(path)) return Promise.resolve();
-    return copyFile(path, this.destPath).then(() => {
+    return copyFile(path, this.destinationPath).then(() => {
       debug(`Copied ${path}`);
       this.updateTime();
       this.error = null;
--- a/test/_test_helper.js
+++ b/test/_test_helper.js
@@ -1,0 +1,57 @@
+const path = require('path');
+const fs = require('fs-extra');
+
+const rootPath = process.cwd();
+const tmp = path.join(require('os').tmpDir(), 'brunch-tests');
+
+const createPackageJson = function() {
+  const minimalJson = `{
+    "name": "brunch-app",
+    "description": "Description",
+    "author": "Your Name",
+    "version": "0.1.0",
+    "dependencies": {},
+    "devDependencies": {
+      "javascript-brunch": "^2.0.0"
+    }
+  }`;
+
+  fs.writeFileSync('package.json', minimalJson);
+};
+
+module.exports.prepareTestDir = function prepareTestDir() {
+  fs.mkdirsSync(tmp);
+  process.chdir(tmp);
+  createPackageJson();
+};
+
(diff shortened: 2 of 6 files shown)


lib/fs_utils/asset.js
```diff
@@ -45,7 +45,7 @@ class Asset {
   copy() {
     const path = this.path;
     if (isIgnored(path)) return Promise.resolve();
-    return copyFile(path, this.destPath).then(() => {
+    return copyFile(path, this.destinationPath).then(() => {
```
goshakkk: oops... 👍

goshakkk: I've actually already fixed this one in https://github.com/brunch/brunch/commit/3f20ff4d3571644794717572c8b75c52a37f770f

Combining a fix and adding some test structure in a single PR is probably not the best thing... at least to get the fix in asap :) That said, I'll look at the tests portion now

test/_test_helper.js
```diff
@@ -0,0 +1,57 @@
...
+
+module.exports.fileContains = function fileContains(t, path, content) {
+  try {
+    t.true(fs.readFileSync(path, 'utf8').includes(content));
+  } catch (e) {
+    t.fail(e.message);
+  }
+};
+
+module.exports.fileNotContains = function fileNotContains(t, path, content) {
```
goshakkk: doesNotContain?

test/build.js
```diff
@@ -0,0 +1,271 @@
...
+        }
+      }
+    };`,
+    'app': {
+      'assets': {
+        'index.html': '<h1>hello world</h1>'
+      },
+      'initialize.js': 'console.log("hello world")'
+    }
+  });
```
goshakkk: Wow, this is actually looking pretty good. I have had an idea to create several brunch projects in the repo and operate on these, but this `fixturify` thing seems to be an even better option 👌

test/build.js
```diff
@@ -0,0 +1,271 @@
...
+});
+
+;require.register("b.js", function(exports, require, module) {
+fileb
+});
+
+;require.register("c.js", function(exports, require, module) {
+filec
+});
+`);
```
goshakkk: This is probably ok, for now, but maybe we shouldn't heavily rely on the exact contents of output file. What will probably make sense further down the line, something that I've been thinking about implementing when I'll get my hands down to improving our testing, is running the resulting js in a headless webkit and very that things indeed work properly (which will also guard us against bugs in commonjs runtime, etc)

But as I said, totally fine for now

goshakkk:
@paulmillr this one looks great to provide some foundation for testing. I'll later work on covering even more of Burnch configurations with tests.

test/_test_helper.js
marcioj: Woops, I'll fix in a future PR

test/build.js
marcioj: This is a good idea. I'll try to give a look when I have some time.

marcioj:
Thanks for the review @goshakkk

Issue #1268, opened 2016-03-31
browsersync, autoreload with brunch and jade - have to reload twice before seeing changes?


alancwoo:
### Description

In a simple [brunch](http://brunch.io) project, I have tried using both browsersync and autoreload to keep my browser updated automatically with changes.

Edits to javascript and css are immediately updated, but when updating a jade template, the browser performs a reload, but I have to save again or manually reload the browser to see the change.

It's strange because according to the debug output, it seems to do two reloads, and also copy the html before the second reload.

**On change to jade template and save:**

    brunch:file-list Reading 'app/index.jade' +7s
      brunch:watch File 'app/index.jade' received event 'change' +0ms
      brunch:pipeline Compiling 'app/index.jade' with 'JadeCompiler' +0ms
      brunch:pipeline getDependencies 'app/index.jade' with 'JadeCompiler' +2ms
      brunch:file-list Compiled file 'app/index.jade'... +1ms
      brunch:write Writing 1/4 files +65ms
      brunch:generate Concatenating app/index.jade, app/js/app.js to public/app.js +0ms
      brunch:common Writing file 'public/app.js' +1ms
    23 Feb 16:11:04 - info: compiled index.jade and 1 cached file into app.js in 70ms
    [BS] Reloading Browsers...
      brunch:watch File 'app/assets/index.html' received event 'change' +32ms
      brunch:file-list Copied asset 'app/assets/index.html' +1ms
      brunch:file-list Copied asset 'app/assets/index.html' +0ms
      brunch:write Writing 0/4 files +67ms
    23 Feb 16:11:04 - info: copied index.html in 69ms
    [BS] Reloading Browsers...

### Environment

1. Brunch: 1.8.5
2. Node: 5.4.0
3. NPM: 3.3.12
4. Operating system: OS X 10.11.3

### `package.json` contents

```
{
  "name": "brunch-jade-test",
  "version": "0.0.1",
  "description": "brunch-jade-test",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "brunch": "^1.8.5",
    "css-brunch": "^1.7.0",
    "jade-brunch": "^1.8.1",
    "javascript-brunch": "^1.8.0",
    "static-jade-brunch": "^1.8.0",
    "stylus": "^0.53.0",
    "stylus-brunch": "^1.8.1"
  },
  "dependencies": {
[...]

PR #1356, opened 2016-05-10
Implement API for static compiler plugins

goshakkk:
Allows to implement plugins like static-jade-brunch much simpler, by
introducing another type of compilers — static compilers. They are
distinct from js/css/template compilers in a few regards:

- first and foremost, there is no 'joining' step for static file — they
  are written directly. Each source file corresponds to one compiled
  file
- allows to handle any plain-text static file. The plugin dictates the
  resulting file extension. (But not binary files. In practice, it means
  there can be an svgo compiler that both takes and produces svg.
  Doesn't apply to png, jpg, and other binary formats though)

A side benefit of this first-class support is that there are no two
successive change events with almost no delay, which was causing an
issue with autoreload sometimes (GH-1268)

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -229,6 +229,35 @@ class MyCompiler {
 
 Note: exported JS will not be compiled or linter by any other plugin and its `require` statements will not be resolved. Make sure your exported JS is self-contained.
 
+### Static file compilers
+
+Sometimes, you would want to process different kinds of files, to which the Brunch's general compile-join-write logic does not apply.
+Jade templates to HTML is one example.
+You want to have a `.jade` file compiled into `.html`.
+Previously, what you would do in this case was to hook into `onCompile` and look for jade files... and then compile them and write them manually. Sucks.
+
+So starting Brunch `<unreleased>`, there is a better way.
+
+```javascript
+class StaticJadeCompilerNew {
+  compile(params) {
+    const path = params.path;
+    const data = params.data;
+
+    return new Promise((resolve, reject) => {
+      toHtml(path, data, (err, data) => {
+        if (err) return reject(err);
+        resolve(data);
+      });
+    });
+  }
+}
+StaticJadeCompilerNew.prototype.brunchPlugin = true;
+StaticJadeCompilerNew.prototype.type = 'static';
+StaticJadeCompilerNew.prototype.extension = "jade";
+StaticJadeCompilerNew.prototype.targetExtension = "html";
+```
+
 ## Publishing
 
 Making your plugin available to everyone is as simple as
--- a/lib/config.js
+++ b/lib/config.js
@@ -47,6 +47,11 @@ const configBaseSchema = v.object({
       if (['app', 'test', 'vendor', 'assets'].indexOf(key) !== -1) {
         return 'was removed, use `config.paths.watched` instead';
       }
+    },
+    specifics: {
+      statics: v.object({
+        rootPath: v.string.default('app')
+      }).default({})
     }
   }, v.object({
     joinTo: v.either(
--- a/lib/fs_utils/source_file.js
+++ b/lib/fs_utils/source_file.js
@@ -134,6 +134,14 @@ class SourceFile {
     this.removed = false;
     this.disposed = false;
     this.compile = makeCompiler(path, this, linters, compilers, wrap, jsWrap, depCompilers);
+
+    if (type === 'static') {
+      const staticComp = compilers[0];
+      this.sourceExt = staticComp.extension;
+      this.sourcePattern = staticComp.pattern;
+      this.targetExtension = staticComp.targetExtension;
+    }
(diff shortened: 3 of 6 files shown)


docs/plugins.md
```diff
@@ -229,6 +229,37 @@ class MyCompiler {
 
 Note: exported JS will not be compiled or linter by any other plugin and its `require` statements will not be resolved. Make sure your exported JS is self-contained.
 
+### Static file compilers
```
goshakkk: @paulmillr this is a glimpse of what `static-jade-brunch` would look like with this new API

lib/config.js
```diff
@@ -47,6 +47,11 @@ const configBaseSchema = v.object({
       if (['app', 'test', 'vendor', 'assets'].indexOf(key) !== -1) {
         return 'was removed, use `config.paths.watched` instead';
       }
+    },
+    specifics: {
```
goshakkk: yay, some flexibility of `skemata` came in very handy, allowing `files.statics` to be different from other `files.` settings

lib/fs_utils/source_file.js
```diff
@@ -134,6 +134,11 @@ class SourceFile {
     this.removed = false;
     this.disposed = false;
     this.compile = makeCompiler(path, this, linters, compilers, wrap, jsWrap, depCompilers);
+
+    if (type === 'static') {
+      this.targetExtension = compilers[0].targetExtension;
+    }
```
goshakkk: this is needed so that we later know which extension to use for the file when writing

lib/fs_utils/write.js
```diff
@@ -107,6 +107,7 @@ const checkWritten = (fileList, files, startTime) => {
     if (file.error) logger.error(formatWriteError(file));
     if (file.compilationTime >= startTime) {
       Object.keys(file.targets).forEach(type => {
+        if (type === 'static') return;
```
goshakkk: There is no `joinTo` for statics so this checker should ignore them entirely

lib/fs_utils/write.js
```diff
@@ -154,4 +174,10 @@ const write = (fileList, config, joinConfig, optimizers, startTime) => {
   })).then(() => Promise.resolve({changed, disposed}));
 };
 
+const write = (fileList, config, joinConfig, optimizers, startTime) => {
+  const statics = writeStatics(fileList, config, startTime);
+  const joins = writeJoins(fileList, config, joinConfig, optimizers, startTime);
+  return statics.then(() => joins);
```
goshakkk: since we now have two different flows for writing, i've split them into several functions. one for joinTo files, one for static files

lib/fs_utils/write.js
```diff
@@ -118,7 +119,26 @@ const checkWritten = (fileList, files, startTime) => {
...
 };
 
-const write = (fileList, config, joinConfig, optimizers, startTime) => {
+const writeStatics = (fileList, config, startTime) => {
+  const allFiles = Array.from(fileList.files.values());
+  const statics = allFiles.filter(f => f.type === 'static' && f.compilationTime >= startTime);
+  const validStatics = statics.map(stat => {
+    const relPath = sysPath.relative(config.files.statics.rootPath, stat.path);
+    if (relPath.slice(0, 2) === '..') {
+      logger.warn(`Resolved ${stat.path} as ${relPath}, skipping. Relative path of a static file should never start with '..'. Possible fix: change 'config.files.statics.rootPath'.`);
```
goshakkk: Yeah. If `files.statics.rootPath` is `app` (default) and there is a matching file in say `test`, its relative path will be `../test/filename` — which is at all undesired, since by joining it with `public`, that we'd obtain would be... `test/filename`, not something in `public`

goshakkk:
After this is merged, we could also update `static-jade-brunch` to take advantage of this.

@ilkosta @paulmillr what do you think?

goshakkk:
(the test is failing because an updated version of deppack wasn't released just yet. it was there before this PR)

goshakkk:
Also, obviously, static-jade-brunch wouldn't require jade-brunch with this. In fact, the presence of the two at the same time won't work out very well (if both are watching the same `jade` extension) — the users should either choose to have static or js templates, or use a different extension for static jade (and configure static-jade-brunch accordingly)

lib/fs_utils/write.js
```diff
@@ -118,7 +119,26 @@ const checkWritten = (fileList, files, startTime) => {
...
+      return;
+    }
+    const relWithExt = relPath.replace(/\.(\w+)$/, '.' + stat.targetExtension);
```
goshakkk: replacing any extension instead of `plugin.extension` should work just fine in case of `pattern` now

paulmillr:
So this handles a simple case for static-jade-brunch. What about other cases? For example, image sprites (many assets -> one output).

PR #1363, opened 2016-05-11
Make watcher restarts clean

goshakkk:
This allows to cleanly restart the watcher. The previous approach with
manually trying to clean require.cache didn't work well and produced
various memory leaks.

This commit changes the brunch init process: the binary does not
directly run brunch commands, but instead, creates a child process that
processed the commands. In case the child process needs to restart, it
kills it and runs another one.

It works properly under two conditions: both global brunch binary and
the local brunch installation contain this commit. If the global brunch
is older, or if the local brunch is older, the old behavior will be
preserved.

---

Note: what this means is there will be at least two processes when running `brunch watch` (but not other commands) — one that handles restarts, another that does the actual processing. However, the overhead of this approach should not be significant since the binary process does not actively do anything but wait for the `restart` message from the watcher. Its memory footprint should also be minimal since it only requires system modules.

### DIFF
--- a/bin/brunch
+++ b/bin/brunch
@@ -5,6 +5,7 @@ global.appStartTime = Date.now();
 
 var sysPath = require('path');
 var fs = require('fs');
+var childProcess = require('child_process');
 var version = process.version;
 var verDigit = parseInt(version.match(/^v(\d+)\./)[1]);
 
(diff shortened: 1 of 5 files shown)


paulmillr:
Let's find a way of not running the second process by default.

goshakkk:
A new process is only spawned only when it's really needed — i.e. only in watcher. No way around this if we want to be able to cleanly restart the watcher without quitting the process and having the user run brunch again manually.

PR #1367, opened 2016-05-15
Additional tests

marcioj:
This PR create some additional tests for `brunch build` and `brunch watch`.

### DIFF
--- a/bin/brunch
+++ b/bin/brunch
@@ -29,6 +29,7 @@ var exists = function(path) {
 };
 
 var runSeparateWatchProcess = function(cliPath) {
+  process.env.BRUNCH_FORKED_PROCESS = true;
   var proc = childProcessFork(cliPath, process.argv.slice(2));
   proc.on('message', function(message) {
     if (message === 'reload') {
--- a/lib/fs_utils/file_list.js
+++ b/lib/fs_utils/file_list.js
@@ -46,6 +46,7 @@ class FileList extends EventEmitter {
     this.copying = new Set();
     this.compiled = new Set();
     this.initial = true;
+    this.disposed = false;
 
     this.on('change', this._change);
     this.on('unlink', this._unlink);
@@ -163,6 +164,7 @@ class FileList extends EventEmitter {
   copy(asset) {
     const path = asset.path;
     const resetCopy = (p) => {
+      if (this.disposed) { return p; }
       this.copying.delete(path);
       this.resetTimer();
       return p;
@@ -198,6 +200,7 @@ class FileList extends EventEmitter {
     } else {
       debug(`Reading ${path}`);
       readFileAndCache(path, error => {
+        if (this.disposed) { return; }
         if (error) throw new Error(formatError('Reading', error));
         // .json files from node_modules should always be compiled
         if (!ignored && (compiler && compiler.length) || deppack.isNpmJSON(path)) {
@@ -237,6 +240,7 @@ class FileList extends EventEmitter {
       this[k] = null;
     });
     this.assets.splice(0);
+    this.disposed = true;
 
     // You're frozen when your heart's not open.
     Object.freeze(this);
--- a/lib/watch.js
+++ b/lib/watch.js
@@ -166,7 +166,10 @@ class BrunchWatcher {
         process.exit(code);
       });
     } else {
-      setTimeout(() => {
+      if (this.nothingToCompileTimer) {
+        clearTimeout(this.nothingToCompileTimer);
+      }
+      this.nothingToCompileTimer = setTimeout(() => {
         if (this.fileList.files.size === 0) {
           logger.warn(nothingToCompile);
         }
@@ -240,12 +243,9 @@ class BrunchWatcher {
 
   restartBrunch(reinstallType) {
     const restart = () => {
-      this.fileList.dispose();
-      this.watcher.close();
-      workers.close();
       // we need this to keep compatibility with global `brunch` binaries
       // from older versions which didn't create a child process
-      if (process.send) {
+      if (process.send && process.env.BRUNCH_FORKED_PROCESS === '1') {
         process.send('reload');
       } else {
         const opts = this._constructorOptions;
@@ -263,6 +263,11 @@ class BrunchWatcher {
         return restart();
       }
     };
+
+    clearTimeout(this.nothingToCompileTimer);
+    this.fileList.dispose();
+    this.watcher.close();
+    workers.close();
 
     const reInstallStep = reinstallType ?
       application.install(this.config.paths.root, reinstallType) :
--- a/package.json
+++ b/package.json
@@ -38,8 +38,8 @@
     "brunch": "./bin/brunch"
   },
   "scripts": {
-    "test": "eslint lib test && ava",
-    "test:coverage": "nyc ava && nyc report --reporter=html"
+    "test": "eslint lib test && ava -s",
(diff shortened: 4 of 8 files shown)


package.json
```diff
@@ -38,8 +38,8 @@
     "brunch": "./bin/brunch"
   },
   "scripts": {
-    "test": "eslint lib test && ava",
-    "test:coverage": "nyc ava && nyc report --reporter=html"
+    "test": "eslint lib test && ava -s",
+    "test:coverage": "nyc ava -s && nyc report --reporter=html"
```
marcioj: Since now we have test/build.js and test/watch.js, by default these files will run serially, but we need to run files sequentially to prevent that both write to the tmp test dir at the same time.

lib/watch.js
```diff
@@ -240,12 +243,9 @@ class BrunchWatcher {
 
   restartBrunch(reinstallType) {
     const restart = () => {
-      this.fileList.dispose();
-      this.watcher.close();
-      workers.close();
       // we need this to keep compatibility with global `brunch` binaries
       // from older versions which didn't create a child process
-      if (process.send) {
+      if (process.send && process.env.BRUNCH_FORKED_PROCESS === '1') {
```
marcioj: Added `BRUNCH_FORKED_PROCESS` env to identify when this file is called from `bin/brunch`. Because ava tests run in a subprocess, `process.send('reload');` was being called, but since there is not any handler for reload, the reload never triggered.

lib/fs_utils/file_list.js
```diff
@@ -163,6 +164,7 @@ class FileList extends EventEmitter {
   copy(asset) {
     const path = asset.path;
     const resetCopy = (p) => {
+      if (this.disposed) { return p; }
```
marcioj: I think there are situations where the FileList is disposed, but some file operations are still running. So when their callbacks are resolved, some properties are null, or read-only because we froze them in the dispose method. Accessing these properties give errors. I added this disposed flag to skip some operations when the object was already disposed.

bin/brunch
```diff
@@ -29,6 +29,7 @@ var exists = function(path) {
 };
 
 var runSeparateWatchProcess = function(cliPath) {
+  process.env.BRUNCH_FORKED_PROCESS = true;
```
goshakkk: Have you tested this will work? I think env can be passed to `child_process` call explicitly

test/_test_helper.js
```diff
@@ -19,10 +19,34 @@ const createPackageJson = function() {
...
+      "node_modules",
+      "bower_components",
+      "test",
+      "tests"
+    ],
+    "dependencies": {},
+    "devDependencies": {}
+  }`;
+
+  fs.writeFileSync('bower.json', minimalJson);
```
goshakkk: Why do we need bower there? Bower is going away and brunch recommends to use npm exclusively for client-side deps.

lib/fs_utils/file_list.js
goshakkk: Could this be indicative of a higher level issues? In general if a file list is disposed, it shouldn't be used anywhere, and nothing should reference it

test/watch.js
```diff
@@ -0,0 +1,441 @@
...
+
+const watch = function(...args) {
+  watcher = brunch.watch(...args);
+};
+
+test.beforeEach.cb((t) => {
+  teardownTestDir();
+  prepareTestDir();
+
+  if (watcher) {
```
goshakkk: should the watcher be closed in `afterEach` instead?

test/watch.js
```diff
@@ -0,0 +1,441 @@
...
+      'initialize.js': 'console.log("hello world")'
+    }
+  });
+
+  var calls = 0;
+
+  const onCompile = function() {
+    calls++;
+
+    if (calls === 1) {
```
goshakkk: Is the conditional needed? I think it can be more linear and straightforward, and less nested, with an assertion with a custom error message instead:

```javascript
t.true(calls === 1, `Unexpected number of calls ${calls}`);
fileExists....
```

test/watch.js
```diff
@@ -0,0 +1,441 @@
...
+      fileExists(t, 'public/app.js.map');
+      fileContains(t, 'public/app.js', '//# sourceMappingURL=app.js.map');
+      fileContains(t, 'public/app.js', `require.register("initialize.js", function(exports, require, module) {
+console.log("hello world")
+});`);
+      fileContains(t, 'public/index.html', '<h1>hello world</h1>');
+
+      fs.writeFileSync('app/new-file.js', 'console.log("new")');
```
goshakkk: Wondering if there's a way to write this test in a more linear fashion, so that it's immediately clear that the following assertions are executed after this new file is written and processed...

test/watch.js
```diff
@@ -0,0 +1,441 @@
...
+      t.end();
+    } else {
+      t.fail(`Unexpected number of calls ${calls}`);
+    }
+  };
+
+  watch({}, onCompile);
+});
+
+test.serial.cb('install bower components if bower.json changes', t => {
```
goshakkk: probably ok to not test bower at all. npm is where it's at

cc @paulmillr

test/watch.js
goshakkk: Perhaps we could use async iterators/generators here, to achieve something like

```javascript
brunch.watch({}, onCompile) // produces the `compilation` generator

await compilation;

// first compile checks
fs.write....

await compilaton;
// second compile checks...
```

bin/brunch
marcioj: It should, but right now is incorrect .... See #1368

test/watch.js
marcioj: I'm testing any feature provided by brunch today. I'm also using an application with brunch and bower, because in the time I started it, brunch didn't have the npm integration and I didn't find a time to migrate. If we want to discourage your usage, I think we should add some warning. Also npm is known to be slower, because this I believe that some people are still using bower, even it having your own issues.

test/watch.js
marcioj: I'm putting everything in `beforeEach` because in this ava version, if a test throws an error the `afterEach` is not called. This is why in the `beforeEach` first is called the `teardownTestDir` Because if in the previous run something failed, we cleanup the test state. Fortunately, ava master branch has a `afterEach.always` which was introduced [in this PR](https://github.com/sindresorhus/ava/pull/806), when it got a new release we can use it and test in a sane way.

lib/fs_utils/file_list.js
goshakkk: It appears to have been caused by first removing the temp dir and only *then* closing the watcher. https://github.com/brunch/brunch/pull/1370 deals with that underlying cause.

test/watch.js
goshakkk: Yeah, it can indeed look somewhat like that using generators (which are supported natively by node 5 & 6, unlike async/await): https://github.com/brunch/brunch/pull/1370

```javascript
  watch({}, function *main(compilation) {
    yield compilation();
    t.true(fs.readdirSync('./node_modules').indexOf('lodash') === -1);

    const packageJson = JSON.parse(fs.readFileSync('package.json', 'utf8'));
    packageJson.dependencies.lodash = '*';
    fs.writeFileSync('package.json', JSON.stringify(packageJson, null, 2));

    yield compilation();
    t.true(fs.readdirSync('./node_modules').indexOf('lodash') !== -1);
    t.end();
  });
```

marcioj: This looks really good!

PR #1368, opened 2016-05-17
Fix BRUNCH_FORKED_PROCESS global variable

marcioj:
I added the BRUNCH_FORKED_PROCESS variable in the previous PR but your usage was incorrect, so the reload message will never trigger. Sorry for that.

### DIFF
--- a/bin/brunch
+++ b/bin/brunch
@@ -29,7 +29,7 @@ var exists = function(path) {
 };
 
 var runSeparateWatchProcess = function(cliPath) {
-  process.env.BRUNCH_FORKED_PROCESS = true;
+  process.env.BRUNCH_FORKED_PROCESS = 'true';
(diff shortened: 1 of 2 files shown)


paulmillr:
fails tests

marcioj:
@paulmillr the error seems something related to npm cache and just failed on one of the 4 jobs. Could you please trigger a new build on travis to see if it passes?

paulmillr:
Still broken @marcioj

marcioj:
I tested here (linux) with both node 5 and 6 and all tests passed. This was also true on Travis, but OSX sometimes fail. I'm unable to test in an OSX right now through. But looking the previous builds node 6 passed on OSX in the first run, but failed on the second. It might be some synchronization issue that happens sometimes.

PR #1369, opened 2016-05-17
Static compilation new take

goshakkk:
Changes to bring static asset compilation as to what @paulmillr and I have discussed.

This PR introduces an entirely new plugin API method: `compileStatic`. (`getDependencies` will then be used for both `compile` *and* `compileStatic` — as dependency discovery logic will be identical for both.)

The static asset compilation now will work on files from `app/assets`, this seems beneficial as it makes clear what will be compiled the regular way, and what will be compiled one-to-one. This means `app/index.jade` in will be a JS template, and `app/assets/index.jade` will turn into `public/index.html` — and both can be actually handled by the same plugin, `jade-brunch`, without having a separate plugin for this purpose.

- [x] tests
- [x] squash

### DIFF
--- a/lib/fs_utils/asset.js
+++ b/lib/fs_utils/asset.js
@@ -4,6 +4,9 @@ const sysPath = require('../path');
 const copyFile = require('quickly-copy-file');
 const prettify = require('../helpers').prettify;
 const isIgnored = require('./common').isIgnored;
+const pipeline = require('./pipeline');
+const plugins = require('../plugins');
+const logger = require('loggy');
 
 // Asset: Simple abstraction on top of static assets that are not compiled.
 
@@ -20,6 +23,12 @@ const getAssetDirectory = (path, convention) => {
   return split.map((part, index) => {
     return split.slice(0, index).concat([part, '']).join(sep);
   }).find(convention);
+};
+
+const replaceExt = (rel, compiler) => {
+  const oldRe = plugins.patternFor(compiler, true);
+  const newExt = compiler.staticTargetExtension;
+  return rel.replace(oldRe, '.' + newExt);
 };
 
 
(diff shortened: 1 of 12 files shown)


lib/fs_utils/asset.js
```diff
@@ -66,4 +74,56 @@ class Asset {
   }
 }
 
+class CompiledAsset {
```
goshakkk: Currently brunch has two notions of file: SourceFile, something that's compiled along with source maps and stuff and then joined into one file; and Asset, which is just copied as-is.

I couldn't find a place for a new type of asset, one that is compiled, in either of these so I created a new class with has the parts it needs from both (without the stuff it does *not* need)

lib/fs_utils/file_list.js
```diff
@@ -41,6 +41,7 @@ class FileList extends EventEmitter {
     this.depCompilers = config.npm.compilers;
 
     this.files = new Map();
+    this.staticFiles = new Map();
```
goshakkk: A separate map to separate compiled assets from source files

lib/fs_utils/file_list.js
```diff
@@ -121,11 +123,17 @@ class FileList extends EventEmitter {
...
+    path = normalizePath(path);
+    return this.staticFiles.get(path);
+  }
+
+  compileDependencyParents(path, isStatic) {
     path = normalizePath(path);
     const paths = [];
     const parents = [];
-    this.files.forEach((dependent, fpath) => {
+    const files = isStatic ? this.staticFiles : this.files;
```
goshakkk: The dependency tracking logic is similar for both, this check as to where pick the files from is about the only thing that differs in this regard

lib/fs_utils/file_list.js
```diff
@@ -190,25 +198,44 @@ class FileList extends EventEmitter {
...
+        if (!isAsset && !ignored && (compiler && compiler.length) || deppack.isNpmJSON(path)) {
           const sourceFile = this.find(path) ||
             this._add(path, compiler, linters, isHelper);
           this.compile(sourceFile);
+        } else if (isStatic && !ignored) {
+          const staticFile = this.findStatic(path) || this._addStatic(path, compiler);
+          this.compile(staticFile);
         }
-        if (!this.initial) this.compileDependencyParents(path);
+        if (!this.initial) this.compileDependencyParents(path, isStatic);
```
goshakkk: This and above, too, is mostly similar for both

lib/fs_utils/generate.js
```diff
@@ -247,6 +248,21 @@ const generate = (path, targets, config, optimizers) => {
     });
 };
 
+generate.writeFile = writeFile;
+
+generate.writeStatics = (toRemove, toWrite) => {
+  const removePromise = toRemove.reduce((promise, file) => {
+    file.dispose();
+    return promise.then(() => unlink(file.destinationPath));
```
goshakkk: So that if `app/assets/index.jade` is deleted we *also* delete `public/index.html`

lib/fs_utils/pipeline.js
```diff
@@ -18,7 +18,7 @@ const prepareError = (type, stringOrError) => {
     stringOrError;
   const error = new Error(string);
   error.code = type;
-  return error;
+  return Promise.reject(error);
```
goshakkk: This is a minor improvement but since the result of this function is always wrapped into a rejection, I thought we could as well do it here

lib/fs_utils/pipeline.js
```diff
@@ -65,6 +65,33 @@ const getDependencies = (data, path, compilerDeps, compiler) => {
...
 
+const compileStatic = (path, compiler) => {
+  let source;
+  return readFromCache(path).then(_source => {
+    source = _source;
+    return processJob(CompileStaticJob, {path, source, plugin: compiler});
+  }).then(data => {
+    return Object.assign({}, data, {source});
+  });
+};
```
goshakkk: We don't need the regular pipeline because there are no linters and only *one* compiler. Also no deppack checks, etc

lib/plugins.js
```diff
@@ -232,7 +233,7 @@ exports.init = (config, onCompile) => {
   speed.profile('Loaded plugins');
   const includes = getPluginIncludes(plugins);
   return Promise.resolve({
-    compilers, linters, includes, teardownBrunch, optimizers,
+    compilers, staticCompilers, linters, includes, teardownBrunch, optimizers,
```
goshakkk: static compilers is a new category of compilers, ones which implement `compileStatic`

lib/plugins.js
```diff
@@ -259,6 +259,14 @@ const _getPattern = plugin => {
   }
 };
 
-exports.isPluginFor = path => {
-  return plugin => _getPattern(plugin).test(path);
+exports.patternFor = (plugin, isStatic) => {
+  if (isStatic) {
+    return _getPattern(plugin.staticExtension || plugin.extension, plugin.staticPattern || plugin.pattern);
```
goshakkk: i.e. for static compilation, we will first try `staticExtension` and `staticPattern` (optional), and only then revert to the regular `extension`/`pattern`

lib/fs_utils/asset.js
```diff
@@ -66,4 +74,56 @@ class Asset {
...
+  constructor(path, publicPath, assetsConvention, compiler) {
+    if (compiler.length !== 1) {
+      throw new Error(`Got ${compiler.length} compilers for static file ${path}, can't proceed`);
```
goshakkk: Since assets are only meant to be processed with one compiler, this will blow up if there ever ends up being more that one... the alternative could be to use the first compiler, and print a warning. @paulmillr what do you think?

paulmillr: warning is ok

goshakkk: ok

PR #1372, opened 2016-05-19
Make tests more reliable

goshakkk:


marcioj:
I think these errors are related to the watcher. 
On OSX, chokidar sets the `usePolling` config to false. So this is why the tests might me passing only on it. If I hardcode the `usePolling` with true on linux, all tests passes. Without these changes they fail with the same error on Travis. And looks like this error is caused because [this process.exit call](https://github.com/brunch/brunch/blob/master/lib/watch.js#L189) which is just triggered when usePolling is false.

goshakkk:
@marcioj I actually couldn't get these to pass on Travis even with `usePolling`

goshakkk:
Looks like the reason for these three failing tests ultimately was the fact that after the watcher restart, `test/watch.js` still had a reference the old `watcher`, not a newer one; the newer one thus was never closed, which actually triggered these failures.

goshakkk:
It's green ✨ 💚

Issue #1378, opened 2016-05-26
Allow config.npm.aliases to explicitly ignore dependency


facboy:
Specifically, brunch is incorrectly deciding that I need 'process' somewhere in my code.  I imagine that it's impossible to correct determine whether it's required or not in all cases, but I know it isn't being used, so I'd like to be able to disable it, perhaps by aliasing it to false.

Alternatively, why not remove the detection and let people explicitly require it if they need it?

paulmillr:
Could you provide us the invalid code?

Also agree. We should have a way of ignoring some packages.

goshakkk:
We do need to automatically insert `process` because node modules can use `process` without explicitly requiring it, and we should not burden the users with dealing with that.

The current detection is pretty basic, to ensure it'll be fast enough. We are simply checking if there is any `process.<something>` in the code — which sadly is bound to have some false positives as well. The other option would be to run everything through a JS parser to make sure `process` wasn't defined locally, but it would make things slower.

For a case like this, I think allowing to set `{process: false}` in `npm.aliases` would indeed be a good idea.

facboy:
I'm afraid I can't because I'm not entirely sure what in our (probably unnecessarily) enormous codebase is triggering it, and I can't make the code publicly available :(.  I don't think we want to make things slower, being able to override is ok.

Though out of curiosity, why is 'process' special?

goshakkk:
Honestly the process shim is so small, it probably doesn't even matter.

It is special because, even though in node you can use all of core modules without requiring, in reality 99.99% of the time it's only `process` that is used this way in npm packages which is why it gets this special handling. Without it, node_modules referencing `process` would just throw an error.

Another special case that we had to account for is that we had a few packages use `new Buffer` without requiring `buffer`, and we handle it as well.

facboy:
well maybe i'm doing something wrong, but it ends up trying to include process from somewhere in /usr/lib/node_modules and because i don't join that anywhere in my config, it doesn't work.

in any case:
https://github.com/brunch/brunch/pull/138
https://github.com/brunch/deppack/pull/34

goshakkk:
The warning is because `explore` added the process him file. It's just a warning though.

What do you mean specifically by "doesn't work"?

Regarding the pull requests, I can barely imagine a scenario for wanting to include process if it's never used, so I guess I would name the config option something like `ignoreAutoProcess` (and set it to default to false `v.bool.default(false)`). 

[...]

PR #1381, opened 2016-05-27
Add 'detectProcess' flag to npm config.

facboy:
For https://github.com/brunch/brunch/issues/1378

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -69,7 +69,8 @@ const configBaseSchema = v.object({
     aliases: v.objects({}, v.string),
     styles: v.objects({}, v.array(v.string)),
     static: v.array(v.string).default([]),
-    compilers: v.array(v.string).default([])
+    compilers: v.array(v.string).default([]),
+    usesProcess: v.bool
   }).default({}),
 
   plugins: v.object({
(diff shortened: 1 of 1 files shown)


paulmillr:
I think we want a more general way of excluding packages. Like: `npm: ignored: ['process']`

goshakkk:
Auto-detected shims are about the only case I can think of, so why be more generic?

facboy:
I actually started out with the generic case but decided to keep things simpler.  I can imagine a scenario when a package might be included by another third-party package and I might want to exclude it, but don't really have a concrete use case.

paulmillr:
Could we change it to `includeProcess`? Let's go with that. @facboy

facboy:
arguably that is less accurate, the flag disables the auto-detection of `process`.  if a package explicitly requires `process` it will still be included, which might surprise somebody if they've put `includeProcess: false`.

goshakkk:
@paulmillr, @facboy is correct — this just toggles auto-detection so `detectProcess` seems more apropriate.

ferhatelmas:
> I think we want a more general way of excluding packages. Like: npm: ignored: ['process']

How can we exclude some packages for prod ? 

For example, let's say we require [why-did-you-update](https://www.npmjs.com/package/why-did-you-update) and enable it for dev. For sure, it's not enabled for prod but it's still bundled into vendor.

goshakkk:
@ferhatelmas we currently don't do that... but yeah, it kind of makes sense, I'll add it to our backlog

PR #1398, opened 2016-06-12
fix <notifications> validation to accept boolean or an array of string

HatemJerbi:
as described in official documentation

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -93,7 +93,7 @@ const configBaseSchema = v.object({
     nameCleaner: v.function.default(path => path.replace(/^app\//, ''))
   }).default({}),
 
-  notifications: v.bool.default(true),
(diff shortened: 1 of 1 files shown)


lib/config.js
```diff
@@ -93,7 +93,7 @@ const configBaseSchema = v.object({
     nameCleaner: v.function.default(path => path.replace(/^app\//, ''))
   }).default({}),
 
-  notifications: v.bool.default(true),
+  notifications: v.either(v.bool.default(true), v.array(v.string).default([])).default(true),
```
goshakkk: We don't need the nested defaults here as they won't be taken into account —only top-level one will be (`true`)

Please change to 

```javascript
v.either(v.bool, v.array(v.string)).default(true)
```

PR #1400, opened 2016-06-12
fix <notifications> validation

HatemJerbi:
to accept boolean or an array of string as described in official documentation


HatemJerbi:
weird travis failure

Issue #1402, opened 2016-06-13
unproperly rejected error in a plugin let compilation promise hanging


maxlath:
### Description

A [breaking change in node-sass made my build break, but apparently without properly rejecting the error](https://github.com/sass/node-sass/issues/1592). Thus brunch's pipeline [promisifyPlugin](https://github.com/brunch/brunch/blob/5442da512504cd7ac483b9b211cf9b4930318e33/lib/fs_utils/pipeline.js#L117) promise is never resolved, letting the failing file never be removed from the `FileList` set of compiling files, thus letting `resetTimer` loop indefinitely without ever triggering the `ready` event.

This is what I understood after a fair amount of digging around but from the user perspective, this is how it looks like (in case anyone comes up with this kind of symptoms):
* start building
* debug mode prints the `brunch:list` events for all the files
* once compilation should be done and you would expect brunch to `write` the result, the allocated CPUs jump to 100% indefinitely until process is killed with nothing being written, without the start of a clue of where that could come from.

### Expected behavior

We can't blame brunch for plugins faults but their could be some warning that things aren't going as expected. I see 2 options there:
* either put a timeout on promises after which those a rejected with an error message specifying the incriminated file
* or, watching differences between `compiling` sets from one resetTimer call to the next, and emitting a warning when the set stays the same for x iterations.

This last option seems the best to me, as it gives the information without altering the plugin compilation autonomy.

As an illustration, this is the output of placing `console.log('compiling set', this.compiling)` in `hasPendingFiles` (above [this line](https://github.com/brunch/brunch/blob/5442da512504cd7ac483b9b211cf9b4930318e33/lib/fs_utils/file_list.js#L90)):

```
[...]
compiling set Set { 'app/modules/general/scss/base.scss' }
compiling set Set { 'app/modules/general/scss/base.scss' }
compiling set Set { 'app/modules/general/scss/base.scss' }
13 Jun 11:59:47 - info: compiling.
compiling set Set { 'app/modules/general/scss/base.scss' }
compiling set Set { 'app/modules/general/scss/base.scss' }
compiling set Set { 'app/modules/general/scss/base.scss' }
compiling set Set { 'app/modules/general/scss/base.scss' }
compiling set Set { 'app/modules/general/scss/base.scss' }
[...]
```
Once I got the name of the file failing to compile, my issue was almost solved.

### Actual behavior
(see description)

### Environment

1. Brunch: 2.8.0
2. Node: 4.4.5
[...]

PR #1403, opened 2016-06-13
warning when a file compilation is not responding

maxlath:
addressing #1402 by checking the list of compiling files and warning when the list stays still for 15 iterations


goshakkk:
I'm tackling the mentioned issue in https://github.com/brunch/brunch/pull/1404

PR #1404, opened 2016-06-13
Print warnings when plugins take too long to process

goshakkk:
Closes GH-1402

### DIFF
--- a/lib/fs_utils/generate.js
+++ b/lib/fs_utils/generate.js
@@ -7,7 +7,7 @@ const fsWriteFile = promisify(require('fs').writeFile);
 const mkdirp = promisify(require('mkdirp'));
 const unlink = promisify(require('fs').unlink);
 const deppack = require('deppack'); // needsProcessing, processFiles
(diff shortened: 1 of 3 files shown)


paulmillr:
What about old and slow machines while doing a big uglifyjs pass on a js file?

goshakkk:
Well, what value do you think will be reasonable?

paulmillr:
Don't know. Just saying. We have for example at least two slow servers in production. They take at least 4x more time vs local compilations.

es128:
How about just print status messages a few times at intervals like 15 secs, 30 secs, then finally fully timeout after a pretty long time. 3 minutes maybe?

paulmillr:
I don't think we should time out at all. There are some cases that prolong the build a lot.

Let's keep it simple.

es128:
Is part of the purpose of this not to prevent unmonitored automated build systems from hanging?

Although I guess they tend to have their own built-in mechanisms for detecting and intervening on runaway processes.

goshakkk:
Ok, I updated to make it just print a warning every 15s without ever rejecting.

PR #1411, opened 2016-06-15
Allow preCompile to run before every compile phase.

### LINKED ISSUE #1406: preCompile not called during watched dir compile passes?
This is on Brunch 2.8.0.  I'm trying to write a plugin/use a preCompile hook to run a compile that doesn't do 1:1 compiling of files to js.  However, it looks like while the preCompile hooks (plugin and otherwise) are called when running `brunch build`, or when watchers are reloaded when `brunch-config.js` changes, if a watched file/dir is change, those hooks don't seem to get called on live recompiles.  Is this expected behavior?

(Side note, preCompile doesn't show up anywhere in the docs.)

In case it matters, I'm doing this in the context of a Phoenix Framework project.

nompute:
This change exposes plugins.preCompilers as a function that generates a new set
of preCompile promises on every compile phase.

This resolves #1406 and returns preCompile hooks to their original behaviour.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -180,38 +180,40 @@ exports.init = (config, onCompile) => {
   const optimizers = plugins.filter(propIsFunction('optimize'));
   const teardowners = plugins.filter(propIsFunction('teardown'));
 
-  // Get plugin preCompile callbacks.
-  const preCompilers = plugins.filter(propIsFunction('preCompile'))
-  .map(plugin => {
-    return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


goshakkk:
Looks great. Have you tested it work?

nompute:
Yes, I'm using the functionality in my project.  My current use of the
hooks are pretty basic, but I tried to make as surgical a change as
possible to minimize potential impact.

nompute:
Gentle nudge - any concerns with getting this merged?

Issue #1413, opened 2016-06-16
Spawning a lot of copies of itself when running 'brunch new' command with wrong args


VitaliyR:
### Description

**brunch new marionettejs test** _(missplased name of the project and skeleton name, used without flag -s)_

Everything is doomed when running wrong command within **brunch new**. brunch tries to create directory recursively and not stopping - can't remove. 

_Solution: **killall node**_

### Expected behavior

It should work just fine - error should appear (and it is appearing) and brunch process exit (status !== 0) and that's it.

### Actual behavior

Brunch spawning itself. Each 'copy' tries to create that dir and install some weird packages.

![ps aux screenshot](http://new.tinygrab.com/c1815eb1bdb0d9d45dadeaefe15f8f22467504452e.png)

[Link to screenshot](http://grab.by/QQJo)

### Environment

1. Brunch: 2.7.7
2. Node: 4.2.1
3. NPM: 2.14.7
4. Operating system: macOS 10.11.5

That's all I have for it. brunch new doesn't support -d flag.

goshakkk:
Wow! This was quite a surprising finding, caused by the use of backticks. I couldn't find a way to reliavle escape them so https://github.com/brunch/brunch/pull/1416 changes them to quotes.

PR #1416, opened 2016-06-16
Prevent spawning a lot of processes when misusing brunch new

goshakkk:
Closes GH-1413

It was caused by the use of backticks inside a logger message. When
spawning a growl notification, these caused the shell to start both of
the wrapped commands, resulting in a loop.

Couldn't find a way to properly escape these, so not using them
altogether for now.


VitaliyR:
@goshakkk yes, surprising indeed. We need to have more behaviour test cases for it.

PR #1427, opened 2016-06-29
refactor & remove deep-assign

shvaikalesh:
1. removed `deep-assign` dependency, extracted reusable `customDeepAssign` to helpers; added `filter` as third attribute
2. added lightweight `Array#includes`, `Object.values` and `Object.entries` shims
3. moved `deepFreeze` to helpers; added optional second param
4. refactored `plugins.js`: more cohesion, pure functions, less additional params and index checking
5. replaced `Promise.resolve` with `return` and `Promise.reject` with `throw` in `Promise#then`
6. added more template strings in more places
7. various code style and type coercion fixes

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,21 +1,18 @@
 'use strict';
+require('./polyfills');
+require('coffee-script').register();
+
 const fs = require('fs');
 const sysPath = require('./path');
 const exec = require('child_process').exec;
 const logger = require('loggy');
 const readComponents = require('read-components');
 const anymatch = require('anymatch');
-const coffee = require('coffee-script');
 const debug = require('debug')('brunch:config');
-const deepAssign = require('deep-assign');
-
-const deppack = require('deppack'); // isNpm
-const loadInit = deppack.loadInit;
+const helpers = require('./helpers');
+const deppack = require('deppack'); // isNpm, loadInit
 const mdls = require('./modules');
-
 const skemata = require('skemata');
-
-coffee.register();
 
 const mediator = {};
 const defaultConfigFilename = 'brunch-config';
@@ -44,7 +41,7 @@ const configBaseSchema = v.object({
   files: v.objects({
     keys: ['javascripts', 'templates', 'stylesheets'],
     warner: key => {
-      if (['app', 'test', 'vendor', 'assets'].indexOf(key) !== -1) {
+      if (['app', 'test', 'vendor', 'assets'].includes(key)) {
         return 'was removed, use `config.paths.watched` instead';
       }
     }
@@ -81,9 +78,9 @@ const configBaseSchema = v.object({
   }, false).default({}),
 
   conventions: v.object({
-    ignored: v.anymatch.default([/[\\\/]_/, /vendor[\\\/](node|j?ruby-.*|bundle)[\\\/]/]),
-    assets: v.anymatch.default(/assets[\\\/]/),
-    vendor: v.anymatch.default(/(^bower_components|node_modules|vendor)[\\\/]/)
+    ignored: v.anymatch.default([/[\\/]_/, /vendor[\\/](node|j?ruby-.*|bundle)[\\/]/]),
+    assets: v.anymatch.default(/assets[\\/]/),
+    vendor: v.anymatch.default(/(^bower_components|node_modules|vendor)[\\/]/)
   }).default({}),
 
   modules: v.object({
@@ -156,12 +153,12 @@ const configSchema = v.merge(
 const validateConfig = config => {
   const res = configSchema(config);
   const fmt = skemata.formatObject(res, 'config');
-  if (fmt && fmt.errors.length > 0) {
+  if (fmt && fmt.errors.length) {
     fmt.errors.forEach(error => {
       logger.error(`${error.path}: ${error.result}`);
     });
   }
-  if (fmt && fmt.warnings.length > 0) {
+  if (fmt && fmt.warnings.length) {
     fmt.warnings.forEach(warn => {
       logger.warn(`${warn.path}: ${warn.warning}`);
     });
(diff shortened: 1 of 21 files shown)


lib/config.js
```diff
@@ -507,47 +484,36 @@ const loadComponents = (config, type) => {
...
-      // Returning default values
+    // Returning default values
     return {components: [], aliases: [], order: []};
   });
 };
 
 const loadNpm = (config) => {
   return new Promise((resolve, reject) => {
+    const paths = config.paths;
+    const rootPath = sysPath.resolve(paths.root);
```
paulmillr: This that may fail also.

lib/config.js
```diff
@@ -589,39 +555,32 @@ const tryToLoad = (configPath, fallbackHandler) => {
   let fullPath;
   let basename = configPath;
 
-  let pkg;
-
   try {
-    pkg = require(sysPath.resolve('.', 'package.json'));
+    var pkg = require(sysPath.resolve('.', 'package.json'));
```
paulmillr: Let's use `let` or `const`. No `var`s.

paulmillr: actually nevernind, var seems to look nicer

lib/fs_utils/asset.js
```diff
@@ -15,20 +15,19 @@ const logger = require('loggy');
...
 // Returns String.
 const getAssetDirectory = (path, convention) => {
-  const sep = sysPath.sep;
   const split = path.split('/');
 
   // Creates list similar to:
   // ['app/', 'app/assets/', 'app/assets/thing/', 'app/assets/thing/item.html/']
   return split.map((part, index) => {
-    return split.slice(0, index).concat([part, '']).join(sep);
+    return split.slice(0, index).concat(part, '').join(sysPath.sep);
```
paulmillr: You sure this is not much slower?

lib/fs_utils/asset.js
```diff
@@ -58,14 +57,12 @@ class Asset {
...
       this.error = null;
-      return Promise.resolve();
     }, err => {
       debug(`Cannot copy ${path}: ${err}`);
       this.updateTime();
       const error = new Error(err);
       error.code = 'Copying';
-      this.error = error;
-      return Promise.reject(this.error);
+      throw this.error = error;
```
paulmillr: Can we separate this in two lines? May be very unusual for outside contributors.

lib/fs_utils/generate.js
```diff
@@ -85,7 +84,7 @@ const concat = (files, path, definitionFn, autoRequire, config) => {
   const processor = (file) => {
     root.add(file.node);
     const data = file.node.isIdentity ? file.data : file.source;
-    if (isJs && data.trim().substr(-1) !== semi) root.add(semi);
+    if (isJs && !data.trim().endsWith(';')) root.add(';');
```
paulmillr: won't this allocate strings over and over? This may be a method that's executed lots and lots of times every sec.

lib/fs_utils/pipeline.js
```diff
@@ -18,37 +18,25 @@ const prepareError = (type, stringOrError) => {
...
-      return Promise.resolve(data);
-    }, error => {
-      if (error.toString().match(warningRe)) {
-        logger.warn(`Linting of ${path}: ${error}`);
-        return Promise.resolve(data);
-      } else {
-        return prepareError('Linting', error);
-      }
     });
+  }, Promise.resolve()).then(() => data, error => {
```
paulmillr: Is this a valid syntax?

lib/fs_utils/pipeline.js
```diff
@@ -204,37 +183,31 @@ exports.pipeline = (path, linters, compilers, fileList, depCompilers) => {
...
     const _path = sysPath.resolve('.', path);
     const selectedLinters = [];
-    let selectedCompilers = [];
-
-    if (depCompilers.length) {
-      selectedCompilers = compilers.filter(comp => {
-        return depCompilers.indexOf(comp.brunchPluginName) !== -1 || comp.brunchPluginName === 'javascript-brunch';
-      });
-    }
+    const selectedCompilers = !depCompilers.length ? [] : compilers.filter(compiler => {
```
paulmillr: I really despise double-negations. Let's shuffle this.

lib/plugins.js
```diff
@@ -167,28 +146,33 @@ exports.init = (config, onCompile) => {
...
     // an active environment?
-    return env === '*' || config.env.indexOf(env) >= 0;
+    return env === '*' || config.env.includes(env);
+  });
+
+  debug('Loaded plugins: ' +
+    plugins.map(plugin => plugin.brunchPluginName).join(', ')
+  );
+
+  const respondTo = key => plugins.filter(plugin => {
```
paulmillr: great.

lib/polyfills.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+if (![].includes) {
+  const includes = function(search) {
+    const array = arguments.length > 1 ? [].slice.call(this, arguments[1]) : this;
+    const is = Number.isNaN(search) ? Number.isNaN : (el => el === search);
+    return [].some.call(array, is);
+  };
+  Object.defineProperty(Array.prototype, 'includes', dontEnum(includes));
+  Array.prototype[Symbol.unscopables].includes = true;
```
paulmillr: symbols supported in node 4?

package.json
```diff
@@ -26,7 +26,7 @@
   ],
   "homepage": "http://brunch.io/",
   "author": "Brunch team (http://brunch.io)",
-  "repository": "https://github.com/brunch/brunch.git",
```
paulmillr: I think there may be some parsers / aggregators of NPM package.json files that won't understand the shortcut. Could we change it back?

lib/config.js
shvaikalesh: If it does, promise will be rejected. No need to do extra `try/catch` and call `reject`.

lib/fs_utils/asset.js
shvaikalesh: `sysPath` will have fast properties and `sysPath.sep` will be loop invariant, so no.

lib/fs_utils/asset.js
shvaikalesh: Sure, and I think we should add eslint rule for that.

lib/polyfills.js
shvaikalesh: Yes, they are, even `Array.prototype[Symbol.unscopables]` is present.

lib/fs_utils/pipeline.js
shvaikalesh: Yes, short arrow functions must be wrapped in parens only if they contain `{` or `,`

lib/fs_utils/generate.js
shvaikalesh: Relatively short strings will be cached by v8. `trim` can be improved: we are slicing whitespaces from both sides, but only interested in last `\S` char. I will replace it with regexp: patterns with anchors are very fast.

package.json
shvaikalesh: Will do,

lib/fs_utils/pipeline.js
shvaikalesh: Will do,

Issue #1425, opened 2016-06-28
getDependencies API inconsistency


jacksonrayhamilton:
According to [the docs](https://github.com/brunch/brunch.github.io/blob/922f72d8269007ece5e608ce03847e51218c2db0/docs/plugins.html#L23), the plugin API's `getDependencies` method accepts `path, data` as parameters. But it actually accepts those parameters [in the opposite order](https://github.com/brunch/brunch/blob/3481e3d80d1c67c29a7618bf1756041def557b14/lib/fs_utils/pipeline.js#L62): `data, path`. Those parameters should probably be documented in the correct order.

Also, the methods `lint`, `compile`, `optimize` each accept a single `file` parameter, which happens to have `path` and `data` as properties. What do you think about changing the API to also pass a `file` to `getDependencies`, for consistency?

paulmillr:
good idea.

PR #1440, opened 2016-07-07
Fix get deps

shvaikalesh:
1. Introduced `promiseReduce` helper.
2. Promisification of plugin methods now is not based on `fn.length`: callback is always passed as last argument and unless method returns a promise, will be relied upon.
3. `safePromise` now reports method name and file path.
4. `compiler.compileStatic` is now called via `safePromise` too.
5. refactored `pipeline.js`.

Closes https://github.com/brunch/brunch/issues/1425


paulmillr:
Good job.

PR #1445, opened 2016-07-13
Add "inline" option for inline source maps.

### LINKED ISSUE #1424: Inline source maps
### Description

Outputting source maps to separate files seems to cause race conditions when watching those files.

### Expected behavior

When using `brunch watch` with Karma (running it in a separate terminal tab via `karma start`), `karma-sourcemap-loader`, and the following `karma.conf.js` configuration:

```js
module.exports = function (config) {
  config.set({
    files: [
      {pattern: 'public/**/*.map', included: false, watched: false},
      'public/app.js'
    ],
    preprocessors: {
      '**/*.js': ['sourcemap']
    },
    frameworks: ['mocha'],
    browsers: ['Chrome']
  });
};
```

Then I expect Karma to load app.js and it's source map whenever I make a change to app.js.

### Actual behavior

Sometimes when I make a change to a file, the source map is loaded successfully. But quite often, when the source map is read, it is blank; I am guessing this is due to a race condition where app.js gets saved before app.js.map is completely written to disk.

The stack trace I get in Karma is as follows:

```
28 06 2016 13:17:38.845:INFO [watcher]: Changed file "/project/public/app.js".
28 06 2016 13:17:38.884:ERROR [karma]: SyntaxError: Unexpected end of JSON input
  at Object.parse (native)
  at sourceMapData (/project/node_modules/karma-sourcemap-loader/index.js:12:29)
  at /project/node_modules/karma-sourcemap-loader/index.js:51:11
  at /project/node_modules/graceful-fs/graceful-fs.js:78:16
  at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:439:3)
```

`SyntaxError: Unexpected end of JSON input` is due to the ".js.map" file's contents being blank.

I am guessing this problem could be avoided by making it possible to include source maps inline in app.js, so external watchers couldn't end up reading another file in the midst of it being written; i.e., we could use source maps of the style `//@ sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW...`

Here is a demonstration of this error: [jacksonrayhamilton/brunch-karma-source-map-demo](https://github.com/jacksonrayhamilton/brunch-karma-source-map-demo)

### Environment

1. Brunch: 2.8.2
2. Node: 6.2.1
3. NPM: 3.9.3
4. Operating system: Mac OS El Capitan (v10.11.4)

### `package.json` contents

```json
{
  "scripts": {
    "start": "brunch watch",
    "test": "karma start"
  },
  "devDependencies": {
    "brunch": "^2.8.2",
    "javascript-brunch": "^2.0.0",
    "karma": "^1.1.0",
    "karma-chrome-launcher": "^1.0.1",
    "karma-mocha": "^1.1.1",
    "karma-sourcemap-loader": "^0.3.7",
    "mocha": "^2.5.3"
  }
}
```

### brunch config contents

```javascript
module.exports = {
  files: {
    javascripts: {
      joinTo: {
        'app.js': /.*/
      }
    }
  }
};
```

jacksonrayhamilton:
Fixes #1424.

### DIFF
--- a/lib/fs_utils/generate.js
+++ b/lib/fs_utils/generate.js
@@ -190,7 +190,7 @@ const optimize = (data, map, path, optimizers, sourceFiles) => {
 
 const getInlineSourceMapUrl = map => {
   const base64String = new Buffer(JSON.stringify(map)).toString('base64');
-  return `data:application/json;base64,${base64String}`;
+  return `data:application/json;charset=utf-8;base64,${base64String}`;
 };
 
 const jsTypes = ['javascript', 'template'];
--- a/lib/fs_utils/generate.js
+++ b/lib/fs_utils/generate.js
@@ -190,7 +190,7 @@ const optimize = (data, map, path, optimizers, sourceFiles) => {
 
 const getInlineSourceMapUrl = map => {
   const base64String = new Buffer(JSON.stringify(map)).toString('base64');
-  return `data:application/json;charset=utf-8;base64,${base64String}`;
+  return `data:application/json;base64,${base64String}`;
 };
 
 const jsTypes = ['javascript', 'template'];
--- a/lib/fs_utils/generate.js
+++ b/lib/fs_utils/generate.js
@@ -188,6 +188,11 @@ const optimize = (data, map, path, optimizers, sourceFiles) => {
   return promiseReduce(optimizers, runOptimizer, initial);
 };
 
+const getInlineSourceMapUrl = map => {
+  const base64String = new Buffer(JSON.stringify(map)).toString('base64');
+  return `data:application/json;charset=utf-8;base64,${base64String}`;
+};
+
 const jsTypes = ['javascript', 'template'];
 
 const generate = (path, targets, config, optimizers) => {
@@ -213,7 +218,9 @@ const generate = (path, targets, config, optimizers) => {
   return optimize(code, map, path, foptim, targets)
     .then(data => {
       if (withMaps) {
-        const mapRoute = config.sourceMaps === 'absoluteUrl' ?
+        const mapRoute = config.sourceMaps === 'inline' ?
+          getInlineSourceMapUrl(data.map) :
+          config.sourceMaps === 'absoluteUrl' ?
           mapPath.replace(config.paths.public, '') :
           sysPath.basename(mapPath);
         const controlChar = config.sourceMaps === 'old' ? '@' : '#';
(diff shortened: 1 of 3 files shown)


lib/fs_utils/generate.js
```diff
@@ -188,6 +188,11 @@ const optimize = (data, map, path, optimizers, sourceFiles) => {
   return promiseReduce(optimizers, runOptimizer, initial);
 };
 
+const getInlineSourceMapUrl = map => {
+  const base64String = new Buffer(JSON.stringify(map)).toString('base64');
```
shvaikalesh: `new Buffer` is deprecated in Node 6, use `Buffer.from`

lib/fs_utils/generate.js
```diff
@@ -188,6 +188,11 @@ const optimize = (data, map, path, optimizers, sourceFiles) => {
...
+  return `data:application/json;charset=utf-8;base64,${base64String}`;
```
shvaikalesh: `charset` is extra here

lib/fs_utils/generate.js
Florian-R: Pretty sure `Buffer.from` is not available (yet) in Node 4

jacksonrayhamilton: Brunch claims to support Node `>= 4`, and `Buffer.from` was only added in `5.10`. So, for backwards compatibility, should I do this?

```js
const base64String = (Buffer.from ? Buffer.from(JSON.stringify(map)) : new Buffer(JSON.stringify(map))).toString('base64');
```

lib/fs_utils/generate.js
jacksonrayhamilton: Are you sure? I don't see a default value for the charset in the [RFC](https://tools.ietf.org/html/rfc2397). I'll admit I only added it because the `inline-source-map` package [also specifies it](https://github.com/thlorenz/inline-source-map/blob/6e089385f363456c7d9ac6fb6fbce4a5516f3382/index.js#L104).

lib/fs_utils/generate.js
shvaikalesh: `Buffer.from` **for strings** was added in `5.10`, this code will cause `TypeError` in Node 4, because `Buffer.from` will expect `TypedArray`.

Lets leave `new Buffer`, I will move `base64` complexity to helper later.

lib/fs_utils/generate.js
shvaikalesh: > might specify a charset parameter

Personally, I have always omitted `charset` for HTML/CSS/JS and never encountered problems due to lack of it.

jacksonrayhamilton:
I removed the `charset` parameter, but afterwards I saw https://github.com/substack/node-browserify/issues/753 which seems to indicate we should keep it... is it OK if I revert #2b74535?

shvaikalesh:
OK, lets keep it.
PR LGTM.

PR #1448, opened 2016-07-17
new watch option "-n, --network"

shreeve:
This gives a quick command line option to set the server's hostname to 0.0.0.0 if needed.


paulmillr:
Could you clarify why this is useful in particular as a CLI option vs config?

Florian-R:
> Could you clarify why this is useful in particular as a CLI option vs config?

I'd liked to have this too sometimes. We don't want to commit this config in our team, but when I want to share a quick POC it's convenient to not to have to edit / revert the config before starting the server.

Agree this is not high priority but nice to have.

shreeve:
Same reason @Florian-R mentioned, which is that it's nice to be able to quickly toggle brunch to be visible to other devices on the local network. Having a quick "-a" option makes this super easy.

shreeve:
Based on input from @paulmillr, I changed the option name

Issue #1454, opened 2016-07-25
Evaluate offline-first options


paulmillr:
https://github.com/GoogleChrome/sw-precache

denar90:
It will be really cool if brunch support offline first approach. I've been thinking about making this option (e.g. `offlineFirst`) always switched on. Why? Offline first became a way how app has to look tomorrow. So working offline it's going to be requirement for apps.

By default brunch using `sw-precache` will cache all files in `public` (destination) folder.
`offlineFirst` also can get all `sw-precache` params and more.
We can kinda add some our configuration options. Like: 
 - `swFileName` - name of generated service worker file.
 - `swAutoInclude` (true/false) - automatically includes SW into `index.html` page or all pages which are in `public` dir.

denar90:
I created [sw-precache-brunch](https://github.com/denar90/sw-precache-brunch) plugin instead.
@Florian-R @shvaikalesh thank you for suggestions 👍

denar90:
@paulmillr can we close this issue?

PR #1459, opened 2016-07-29
Add offline first approach

denar90:
This issue is related to #1454 

@paulmillr if we agree on this API I can write some doc for offline first approach.

### DIFF
--- a/lib/fs_utils/write.js
+++ b/lib/fs_utils/write.js
@@ -6,6 +6,7 @@ const logger = require('loggy');
 const deppack = require('deppack'); // getAllDependents
 const formatError = require('../helpers').formatError;
 const generate = require('./generate');
+const SWPreCache = require('sw-precache');
 
 // For use in `.filter()`.
 const changedSince = startTime => generated => {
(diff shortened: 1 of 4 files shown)


Florian-R:
@denar90 Out of curiosity, why did you chose to add another option? Won't this be a better fit for a plugin?

lib/fs_utils/write.js
```diff
@@ -162,6 +171,8 @@ const write = (fileList, config, joinConfig, optimizers, startTime) => {
   })).then(() => {
     return writeStatic(fileList, config, startTime);
   }).then(() => {
+    if (config.offlineFirst.enabled) return writeSWPreCache(config);
```
shvaikalesh: Is there a need to wait for `writeStatic` to complete before calling `writeSWPreCache`? Can we do it in parallel?

denar90:
@Florian-R It can be as separate plugin, but the reason why I propose to use it as part of brunch is offline approach for building applications. As I see the future, app will have to support offline approach be default. So having this behavior in core of brunch will be huge + for it. But I can be wrong and to hurry about using this approach so early in core. If you guys suggest just do it in scope of plugin it's fine by me. Just want to do what is better 😄

shvaikalesh:
Brunch is opinionated build system, but having `GoogleChrome/sw-precache` built right in is too much. And it adds couple dependencies to Brunch itself and prolongs `npm i`. Separate plugin would be awesome.

@denar90, thank you for contribution.

PR #1461, opened 2016-07-30
Normalize/sanitize config.paths with trailing slashes

herenow:
Our internal code expects that the `config.paths.public` and
`config.paths.watched` paths don't end with a trailing slash.

For example, in `lib/fs_utils/generate.js` line 213, we have this:

```js
const len = config.paths['public'].length + 1;
const joinKey = path.slice(len);
```

If we had a trailing slash in the end of the public path, the joinKey
might be different then what we expect.

Another example, in `lib/config.js` line 185, we have:

```js
const isTargetWatched = paths.watched.some(path => target.startsWith(`${path}/`));
```

It appends a trailing slash to the path.

***Note***

I don't think this is the best solution, but given that I'm not 100%
familiar with brunch's source code, and there may be more instances of
code that doesn't expect a trailing slash, sanitizing user input seems to be
best for now.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -289,6 +289,18 @@ const normalizeConfig = config => {
   return config;
 };
 
+const trimTrailingSlashes = config => {
+  const trailingSlash = /\/$/;
+
+  config.paths.public = config.paths.public.replace(trailingSlash, '');
+
+  config.paths.watched.forEach((value, index) => {
+    config.paths.watched[index] = value.replace(trailingSlash, '');
+  });
+
+  return config;
+};
+
 const addDefaultServer = config => {
   if (config.server.path) return config;
   try {
@@ -461,6 +473,7 @@ const loadConfig = (persistent, options, fromWorker) => {
     .then(config => applyOverrides(config, params.env))
     .then(config => helpers.deepAssign(config, params))
     .then(normalizeConfig)
+    .then(trimTrailingSlashes)
     .then(fromWorker || addPackageManagers)
     .then(helpers.deepFreeze(dontFreeze));
 };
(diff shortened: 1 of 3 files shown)


herenow:
I found that whenever I set a path with a trailing slash brunch behaves unexpectedly.

Currently, there are only two cases I know of, but there may be more.

1) [lib/fs_utils/generate.js:198](https://github.com/brunch/brunch/blob/e37095c072e3f5f9587afba250c33d46eafd4aa4/lib/fs_utils/generate.js#L198)

We strip the `public` path from the given `path` to use it as the key for the `entryPoints` config.

2) [lib/config.js:185](https://github.com/brunch/brunch/blob/e37095c072e3f5f9587afba250c33d46eafd4aa4/lib/config.js#L185)

We append a slash to the path, making the `path` and `target` never match.

PS: Feel free to consider this PR as just an issue :)

lib/config.js
```diff
@@ -19,6 +19,24 @@ const BrunchError = require('./error');
 const normalizeChecker = require('anymatch');
 const defaultConfigFilename = 'brunch-config';
 
+const sanitizeConfig = config => {
```
shvaikalesh: Maybe we can come up with less general name than `sanitizeConfig`? `trimTrailingSlashes`?

herenow:
@shvaikalesh I've renamed the function to `trimTrailingSlashes`, I've also removed some unnecessary conditions and moved the  `trimTrailingSlashes` down the promises chain.

shvaikalesh:
@herenow cool, LGTM.

paulmillr:
travis still fails

herenow:
@paulmillr Fixed it

PR #1493, opened 2016-08-25
Fix check deps

shvaikalesh:


shvaikalesh:
Travis on OS X is rarely available lately :(

PR #1535, opened 2016-10-22
Fix const reassign & throw BrunchError

shvaikalesh:
Follow-up of #1533


.eslintrc.json
```diff
@@ -135,6 +135,7 @@
     "constructor-super": 2,
     "generator-star-spacing": [2, "after"],
     "no-class-assign": 2,
+    "no-const-assign": 2,
```
shreeve: nice!

PR #1571, opened 2016-11-28
Architecture bump

paulmillr:


lib/fs_utils/asset.js
```diff
@@ -1,119 +1,91 @@
...
-    return split.slice(0, index).concat(part, '').join(sysPath.sep);
-  }).find(convention);
+// Returns all parent directories:
+// < parentDirs('app/assets/thing/index.html')
+// > ['app/', 'app/assets/', 'app/assets/thing/']
+const parentDirs = path => {
+  return path.split('/').map((part, index, parts) => {
+    return parts.slice(0, index).concat(part, '').join('/');
+  }).slice(0, -1);
 };
```
yavorsky: What about avoiding `return`s:

```
const parentDirs = path =>
  path.split('/').map((part, index, parts) => (
    parts.slice(0, index).concat(part, '').join('/')
  )).slice(0, -1);
```

shvaikalesh: Hey Artem, thanks for the review.

Indeed, we can do that, however I prefer `{}` and `return` even for single-expression arrows that do not "fit" on one line, because it takes less effort to modify then.

Consider the diff for adding statement before return:

```diff
-  return path.split('/').map((part, index, parts) => (
+  return path.split('/').map((part, index, parts) => {
+    if (index === 42) throw new RangeError();
-    parts.slice(0, index).concat(part, '').join('/')
+    return parts.slice(0, index).concat(part, '').join('/');
-  )).slice(0, -1);
+  }).slice(0, -1);
```

versus just

```diff
  return path.split('/').map((part, index, parts) => {
+   if (index === 42) throw new RangeError();
    return parts.slice(0, index).concat(part, '').join('/');
  }).slice(0, -1);
```

shvaikalesh:
Depends on https://github.com/brunch/deppack/pull/46

shvaikalesh:
Node 6 green

Issue #1509, opened 2016-09-15
Config file location is not configurable


otoinsa:
### Description

First of all - I like brunch very much :)
Finally a tool that does it with pretty much no headache :D That's quite the achievement.

I usually keeep all my config files in a config/  folder; the cli tools I use
allow to pass a -c flag usually, where I can define the path of my config.

I think this would be a great change, as it would allow to clean up the code directory even more. 🎱 
### Expected behavior

`brunch watch -c config/brunch-config.js`
Could instruct brunch of where the config location is.
### Actual behavior

`brunch watch -c ./config/brunch-config.js`
won't work

`brunch watch -e ./config/brunch-config.js`
won't work either
### Environment
1. Brunch: 2.8.2
2. Node: 6.5.0
3. NPM: 3.10.3
4. Operating system: OS X 10.11.6
### `package.json` contents

```
{
  "name": "your-app",
  "description": "Description",
  "author": "Your Name",
  "version": "0.0.1",
  "repository": {
    "type": "git",
    "url": ""
  },
  "scripts": {
    "start": "brunch watch",
    "prod": "brunch build --production"
  },
  "dependencies": {
    "react": "~15.3.0",
    "react-dom": "~15.3.0"
  },
  "devDependencies": {
    "auto-reload-brunch": "^2.0.0",
    "babel-brunch": "~6.0.0",
    "babel-preset-es2015": "~6.9.0",
    "babel-preset-react": "~6.11.1",
    "brunch": "^2.4.0",
    "clean-css-brunch": "^2.0.0",
    "css-brunch": "^2.0.0",
    "javascript-brunch": "^2.0.0",
    "uglify-js-brunch": "^2.0.0"
  }
}

```
### brunch config contents

```
module.exports = {
  files: {
    javascripts: {
      joinTo: {
        'vendor.js': /^(?!app)/,
        'app.js': /^app/
      }
    },
    stylesheets: {joinTo: 'app.css'}
  },

  plugins: {
    babel: {presets: ['es2015', 'react']}
  }
};

```

valmirjunior0088:
I would also very much like this feature, since I would also like to load my Brunch configuration file from a "/config" folder.

Perhaps a command line argument or a key in package.json to indicate the path?

stelmakh:
@otoinsa @valmirjunior0088 Hi guys, just wanted to let you know that the feature is in master now and will be available as soon as the new version of Brunch is released. Thanks for your request!

[...]

PR #1583, opened 2016-12-11
Allow usage of --config param

stelmakh:
As requested in #1509, let's allow user to specify a path to their brunch-config file.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -48,7 +48,7 @@ program.command('build [path]')
   .option('-p, --production', 'same as `--env production`')
   .option('-d, --debug [pattern]', 'print verbose debug output to stdout')
   .option('-j, --jobs [num]', 'parallelize the build')
-  .option('-c, --config [path]', 'specify a path to brunch config file')
+  .option('-c, --config [path]', 'specify a path to Brunch config file')
   .action(commands.build);
 
 program.command('watch [path]')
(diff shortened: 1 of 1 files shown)


lib/cli.js
```diff
@@ -59,6 +60,7 @@ program.command('watch [path]')
   .option('-P, --port [port]', 'if `server` was given, listen on this port')
   .option('-d, --debug [pattern]', 'print verbose debug output to stdout')
   .option('-j, --jobs [num]', 'parallelize the build')
+  .option('-c, --config [path]', 'specify a path to brunch config file')
```
shvaikalesh: and here. Otherwise LGTM, thanks for the PR.

yavorsky:
Why this was removed previously?

lib/cli.js
```diff
@@ -48,6 +48,7 @@ program.command('build [path]')
   .option('-p, --production', 'same as `--env production`')
   .option('-d, --debug [pattern]', 'print verbose debug output to stdout')
   .option('-j, --jobs [num]', 'parallelize the build')
+  .option('-c, --config [path]', 'specify a path to brunch config file')
```
stelmakh: Sure thing

stelmakh:
As was said in the deprecation message, it has been replaced with -e option that specifies the environment to use. Probably the --config option has been used mainly for different environments config files rather than to specify the path to config itself.

Issue #1592, opened 2016-12-19
Brunch -d not giving any result


urosgruber:
### Description

After upgrade from 2.8.0 to 2.9.x command `brunch build -d`gives only end response without any debug messages. Looks like debug is in progress because of long build time, but nothing is visible on the screen. Not sure if something is missing in the docs about debug behaviour.

### Expected behavior

Brunch should spit out a whole bunch of debug data

### Actual behavior

`19 Dec 13:11:16 - info: compiled 548 files into 5 files, copied 295 in 86.8 sec`

### Environment

1. Brunch: 2.9.2
2. Node: v5.12.0
3. NPM: 3.8.6
4. Operating system: osx

denysdovhan:
Hey, thank you for reaching out!

I've noticed the same behavior. Gonna try to fix that as soon as it possible.

denysdovhan:
After digging in sources, I've found that `--debug` option is broken after https://github.com/brunch/brunch/commit/34bab856ba82ce8d3700b54bf695f2e87098171b.

urosgruber:
Yes. I can confirm this as well, this is commit where it stop working properly.

denysdovhan:
Fixed in #1596

PR #1595, opened 2016-12-22
Debug hotfix (#1592)

denysdovhan:
This PR introduces a new util called `debug`. It enables `debug` for every module — that means debug messenges will always be shown (which isn't cool).

At this moment, I have no idea why `debug` isn't enables automatically.


denysdovhan:
#1596 provides a better solution.

PR #1597, opened 2016-12-22
Add comment for #1596

denysdovhan:
Suggested in https://github.com/brunch/brunch/pull/1596#issuecomment-268904719


shvaikalesh:
ES6 modules are coming to Node. Imports should be top-level. This should be fixed elsewhere. OK for now.

PR #1600, opened 2016-12-23
Fix rethrow on strings

shvaikalesh:
Replaces https://github.com/brunch/brunch/pull/1577


lib/utils/error.js
```diff
@@ -2,9 +2,14 @@
...
-    .replace(/#{\s*(\w+)\s*}/g, (_, key) => params[key] || '')
-    .concat(`\n${params.error || ''}`);
+  let withError = false;
+  let message = template.replace(/#{\s*(\w+)\s*}/g, (_, key) => {
+    if (key === 'error') withError = true;
+    return params[key] || '';
+  });
+
+  if (!withError) message += `\n${params.error || ''}`;
+  return message;
```
denysdovhan: Why not …?

```js
const format = (template, params) => {
  let withError = false;
  return template
    .replace(/#{\s*(\w+)\s*}/g, (_, key) => {
      if (key === 'error') withError = true;
      return params[key] || '';
    })
    .concat(`\n${!withError ? (params.error || '') : ''}`)
};
```

denysdovhan: Hmm, actually ```.concat(`\n${!withError ? (params.error || '') : ''}`)``` looks not so obvious.

lib/utils/error.js
```diff
@@ -14,16 +13,11 @@ const format = (template, params) => {
...
 class BrunchError extends Error {
   constructor(code, params) {
-    const message = code in messages ?
-      format(messages[code], params || {}) :
-      pipelineMessages.includes(code) ? params.error.message : code;
-
+    const message = format(messages[code] || code, params || {});
     super(message);
 
-    Error.captureStackTrace(this, this.constructor);
```
denysdovhan: Why don't we capture stack trace here?

shvaikalesh: because we `extends`

lib/utils/error.js
shvaikalesh:
```js
`\n${!withError ? (params.error || '') : ''}`
```

meh

lib/utils/error.js
denysdovhan: right, this should work automatically

lib/utils/error.js
shvaikalesh: When people see `.concat` they usually think of arrays, not strings. So I would prefer `+=`.

Issue #1603, opened 2016-12-29
Notifications: Shell injection vulnerability! (was: Problems with $-sign)


lydell:
### Description

In brunch notifications, words starting with a `$` seem to be missing.

### Expected behavior

Words starting with a `$` should be shown.

### Actual behavior

![screenshot from 2016-12-29 13-04-07](https://cloud.githubusercontent.com/assets/2142817/21543589/3641b16c-cdc8-11e6-934b-e6e34ad1bffa.png)

In the above screen shot, note how the word "$black" is missing in the notification. Compare with the text in the terminal.

Here's the test repo: https://github.com/lydell/brunch-notification-bug

1. `npm install`
2. `brunch build`

### Environment

1. Brunch: 2.9.1
2. Node: 6.9.1
3. NPM: 3.10.8
4. Operating system: Ubuntu Gnome 16.04

### `package.json` contents

```json
{
  "private": true,
  "devDependencies": {
    "brunch": "^2.9.1",
    "sass-brunch": "^2.9.0"
  }
}
```

### brunch config contents

```javascript
module.exports = {
  files: {
    stylesheets: {
      joinTo: 'app.css'
    }
  }
}
```

beznosd:
Hello @lydell 

I think the problem is on the OS level.
Will be fixed soon.

lydell:
While commit afefc5e7643f37918ed037e3862461624e483425 fixes my particular case, it is not the proper fix. I've traced down the problem to tj/node-growl#60. The `growl` package fails to do proper escaping.

Here's an example on how to do shell injection with brunch:

0. Clone this repo: https://github.com/lydell/brunch-notification-bug
1. `git checkout shell-injection`
2. Run the below:

```
$ ls u
ls: cannot access 'u': No such file or directory
$ brunch b
31 Dec 12:59:47 - error: Compiling of app/test.js failed. SyntaxError: app/test.js: Bad character escape sequence (1:9)
> 1 | `touch \u`
    |          ^
  2 |  
31 Dec 12:59:47 - info: compiled in 674 ms
$ ls u
u
```

Note how the file `u` was created!

Commit afefc5e7643f37918ed037e3862461624e483425 should be reverted, and the `growl` package should be updated to do proper shell escaping.

beznosd:
Thank you for your notice and great example @lydell 

Yes, we understand that the root of problem is an improper shell escaping of `growl` package, it's just the hot fix to prevent hiding of symbols after $ sign.

[...]

PR #1612, opened 2016-12-30
Fix for issue #1603. Fixed bug with $ sign in notifications.

beznosd:
$ sign was replaced with another one $ sign
https://github.com/brunch/brunch/issues/1603


paulmillr:
Nice kostyl'!

PR #1608, opened 2016-12-30
Yarn support

paulmillr:

PR #1630, opened 2017-01-18
Do not overwrite the NODE_ENV environment variable

IvanSanchez:
The changes in #1608 introduced a bug where brunch will overwrite any user-set value for the `NODE_ENV` environment variable.

For example, the following shell script...

```
export NODE_ENV=staging
node_modules/.bin/brunch build
```

...will build the project with `development` settings, even when I have explicitly set the `NODE_ENV`.

This effectively breaks using more than two environments. I use brunch in conjunction with [the `config` NPM module](https://www.npmjs.com/package/config), and with the current brunch code I cannot build with `staging` settings.

The fix is two-fold:
* Avoid setting a value for `NODE_ENV` if `NODE_ENV` already has a value
* Force the `--production` option if `NODE_ENV` is already set to `production` (e.g. `export NODE_ENV=production; brunch build`)

### DIFF
--- a/lib/utils/config.js
+++ b/lib/utils/config.js
@@ -485,8 +485,8 @@ const initParams = (persistent, options) => {
     stdin: options.stdin != null,
     env: env ? env.split(',') : [],
   };
-  if (options.production == null) {
-    process.env.NODE_ENV = 'development';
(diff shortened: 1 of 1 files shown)


shvaikalesh:
Hey @IvanSanchez, thanks for the fix. We will release patch soon.

Issue #1641, opened 2017-01-22
Vendor pattern ignoring other vendor directories 


mcfarljw:
### Description

I noticed when upgrading from 0.9.x to 0.10.x vendor directories outside of the root directory are not getting compiled. It is possible to get around this issue by manually declaring the vendor conventions. I'm not sure if new behaviors is intended or not.

### Expected behavior

I would expect all vendor directories like `vendor` and `core/vendor` to be handled the same as in previous versions.

### Actual behavior

Only the root directory vendor directory gets compiled and all others are ignored.

### Environment

1. Brunch: 0.10.2
2. Node: 6.9.4
3. NPM: 3.10.10
4. Operating system: macOS 10.12.2

shvaikalesh:
Hey, @mcfarljw. Thanks for the issue and sorry for the trouble. `vendor` pattern was intentionally made more specific in `2.10`. It still seems like a good idea for me, but not for `2.x.` time. Use `master` for now.

PR #1643, opened 2017-01-22
Remove ^ from vendor regexp. Close #1593

### LINKED ISSUE #1593: Problems if `node_modules/` is a symlink
### Description

brunch gets problems if `node_modules/` is a symlink.

### Expected behavior

The build should have produced the exact same output as when `node_modules/` was _not_ a symlink.

### Actual behavior

brunch generates syntactically valid JavaScript, but the JavaScript causes runtime errors. More details below.

### Environment

1. Brunch: 2.9.1
2. Node: 6.9.1
3. NPM: 3.10.8
4. Operating system: Ubuntu 16.04

### `package.json` contents

```json
{
  "private": true,
  "devDependencies": {
    "brunch": "^2.9.1",
    "javascript-brunch": "^2.0.0"
  },
  "dependencies": {
    "jquery": "^3.1.1"
  }
}
```

### brunch config contents

```javascript
module.exports = {
  files: {
    javascripts: {
      joinTo: 'app.js'
    }
  },
  conventions: {
    // This is the default `conventions.vendor` value. If this line is commented
    // out, the problems strangely go away! However, I need to set this to
    // something else in my real project.
    vendor: /^(bower_components|node_modules|vendor)\//,
  },
}
```

### Other useful files, when present (log, `bower.json` etc.)

app/app.js:

```js
require('jquery')
```

### Steps to reproduce

0. Put the above files in a directory and `cd` into it.
1. `npm install`
2. `mv node_modules ../node_modules`
3. `ln -s ../node_modules/ node_modules`
4. `brunch b`

Look at this part of `public/app.js`:

```js
require.register("node_modules/process/browser.js", function(exports, require, module) {

require.register("process/browser.js", function(exports, require, module) {
  require = __makeRelativeRequire(require, {}, "process");
```

Note how there is a `require.register` inside another `require.register`! That causes the following runtime error:

```
Error: Cannot find module 'process' from '/'
```

### Steps to reproduce, variant two

0. Put the above files in a directory and `cd` into it.
1. `npm install`
2. `mv node_modules _node_modules`
3. `ln -s _node_modules/ node_modules`
4. `brunch b`

Output:

```
20 Dec 10:17:50 - error: Processing of _node_modules/process/browser.js failed. Cannot find module 'package.json' 
20 Dec 10:17:50 - info: compiled 2 files into app.js in 634 ms
```

### Why symlink?

The idea was to have the following deploy script:

1. git reset the repo
2. Has package.json changed, or is `../node_modules` missing?
   1. `npm install`
   2. `mv node_modules ..`
3. `ln -s ../node_modules node_modules`
4. `brunch build --production`

This way, I don’t have to `npm install` on every single deploy, which speeds up deploys quite a bit.

denysdovhan:
Because changes #1631 has been reverted. Thx @lydell.

### DIFF
--- a/lib/utils/config-validate.js
+++ b/lib/utils/config-validate.js
@@ -74,7 +74,7 @@ const configBaseSchema = v.object({
   conventions: v.object({
     ignored: v.anymatch.default([/\/_/, /vendor\/(node|j?ruby-.*|bundle)\//]),
     assets: v.anymatch.default(/assets\//),
-    vendor: v.anymatch.default(/(^bower_components|node_modules|vendor)\//),
+    vendor: v.anymatch.default(/(bower_components|node_modules|vendor)\//),
   }).default({}),
 
   modules: v.object({


lydell:
But @shvaikalesh must have done this for a reason: https://github.com/brunch/brunch/commit/a1461dcc316a4c1ca5bc6f8c56177627595c2c82 #1641

denysdovhan:
@lydell I've just removed `^` out of the vendor regexp. There are too many problems with that regexp.

shvaikalesh:
@lydell, thanks for noticing. However, I'm not sure changes were reverted: `^node_modules` is regexp caused issues, and #1631 fixed that. `^vendor` caused issues too, and https://github.com/brunch/brunch/commit/a1461dcc316a4c1ca5bc6f8c56177627595c2c82 fixed that. The current regexp `/(^bower_components|node_modules|vendor)\//` (note `^` after `(` will allow nested `vendor` (I don't think this a good idea, rather a wtf, but if it breaks people let's leave it for `2.x`) and nested `node_modules` (deppack bug).

shvaikalesh:
So this PR is not necessary, let's not change `vendor` any more, who knows what this change will break (I strongly believe `^bower_components` was there for a reason)

lydell:
Ok, I might have been mislead by your commit message: “Bring back old vendor pattern”. I guess that you did that change manually instead of `git revert` to be able to support everything at the same time? I can’t remember which regex is what now ;)

shvaikalesh:
The commit that changed pattern in `2.10` was squashed in #1571, and then 2 commits (#1631 and https://github.com/brunch/brunch/commit/a1461dcc316a4c1ca5bc6f8c56177627595c2c82) reverted it. So I don't think `git revert` was an option, and GitHub edits are so fast! I will add a comment to reflect changes to https://github.com/brunch/brunch/commit/a1461dcc316a4c1ca5bc6f8c56177627595c2c82.

denysdovhan:
@shvaikalesh ok, up to you.

Issue #1648, opened 2017-01-25
[2.10.x] Chains of plugins has issues


devel-pa:
To fix the issue looks like keeping the original file name in the compile parameter will do.

https://github.com/brunch/less-brunch/pull/43

https://github.com/brunch/postcss-brunch/pull/39

https://github.com/brunch/sass-brunch/pull/135

If chained from brunch-less at 
https://github.com/brunch/postcss-brunch/blob/master/index.js#L55
`file.path` is the new CSS file that in fact doesn't exists and should be the original LESS one.

shvaikalesh:
Related: https://github.com/brunch/brunch/issues/1635

devel-pa:
Indeed, hopefully it will solve all the other issues, and I'm pretty sure is related to Brunch, not with plugins.
I'll bet with this, too:
https://github.com/brunch/postcss-brunch/issues/37

devel-pa:
@shvaikalesh I'm searching for a solution and I realize that I can't make any difference between `css-brunch` and `postcss-brunch`, but if I'm removing `css-brunch` it's working as expected. Those two not working together?

shvaikalesh:
I suppose because of source maps. `css-brunch` will be deprecated soon. Brunch `2.10` works with CSS without it. You can safely remove it.

devel-pa:
Then I'll prepare a PR with a code example

PR #1650, opened 2017-01-25
Send inside the plugin the original path when target is not same as the source

devel-pa:
#1648 

I hope it didn't affect other plugins. 
Tested with `less-brunch` (with targetExtension) and `postcss-brunch`.
Remove `css-brunch` as useless.

### DIFF
--- a/lib/fs_utils/pipeline.js
+++ b/lib/fs_utils/pipeline.js
@@ -112,9 +112,12 @@ const compile = (file, compiler) => {
     return deps.then(dependencies => {
       const path = compiled.path || file.path;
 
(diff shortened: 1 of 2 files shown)


shvaikalesh:
I would like to not introduce a new API, but rather join source maps in Brunch, not plugins.

devel-pa:
It is not only about source maps. It's about the fact that postcss doesn't work at all in the current state.

devel-pa:
Really, for production I don't give a damn on the source maps but on browser support.

paulmillr:
keeping this open until we fix postcss etc

devel-pa:
I don't think is a postcss issue, I think is the fact that postcss needs more info.

Issue #1647, opened 2017-01-25
[2.10.3] Handlebars runtime missing.


devel-pa:
Not added to vendor.js at all. Previous versions are OK.

Extra: All versions after 2.8.2 (including plugins) have issues that makes them unusable.

devel-pa:
Probably a hint in case it's the issue: Windows

shvaikalesh:
Hmm, that will help a lot. I think I've got an idea on a fix. Thanks.

devel-pa:
I've tried something, but didn't worked.

denar90:
I've tried to run on windows 10 using latest master of `brunch` and `handlebars-brunch`, works totally fine.
@devel-pa can you share with us config and your `package.json`?

devel-pa:
Hi,
I can't share everything, but:

all dependencies:
```
"devDependencies": {
    "auto-reload-brunch": "~2.7.1",
    "autoprefixer": "6.7.0",
    "brunch": "2.8.2",
    "chai": "^3.5.0",
    "clean-css-brunch": "~2.0.0",
    "css-brunch": "~2.6.1",
    "express": "*",
    "glob": "^5.0.5",
    "handlebars-brunch": "~2.0.0",
    "http-proxy": "*",
    "javascript-brunch": "~2.0.0",
    "jshint-brunch": "~2.0.0",
    "json-brunch": "~1.5.4",
    "less-brunch": "~2.7.1",
    "mocha": "2.5.3",
    "mocha-phantomjs": "4.1.0",
    "postcss-brunch": "0.5.0",
    "replace-brunch": "2.0.0",
    "sinon": "1.17.6",
    "sinon-chai": "^2.8.0",
    "uglify-js-brunch": "~2.0.1"
  }
```

and have in vendor something like this:

```
Handlebars.registerHelper('t', function(key, options) {
});
```

So yes, probably I had to mention that. Sorry

devel-pa:
Of course, this stack doesn't exist anymore as I revert it back to 2.8.2 and company

denar90:
You should use `"handlebars-brunch": "~2.10.0"`. It's kinda latest one https://github.com/brunch/handlebars-brunch/releases/tag/2.10.0
Can you provide use with results of using it?

devel-pa:
Already told you that I'm using that version.
But for the sake of testing, this is the result:
`ReferenceError: Can't find variable: Handlebars`
Same as when I submitted the issue.

[...]

PR #1656, opened 2017-01-28
Rewrite plugin.include handling

shvaikalesh:
Closes https://github.com/brunch/brunch/issues/1647


bin/brunch
```diff
@@ -69,7 +60,7 @@ var loadGlobalBrunch = function() {
 };
 
 var localPath = join(sysPath.resolve('.'), 'node_modules', 'brunch', 'lib', files.old);
-if (exists(localPath)) {
+if (fs.existsSync(localPath)) {
```
denysdovhan: Isn't it dangerous to use `*Sync` method outside of `try..catch`?

bin/brunch
```diff
@@ -44,7 +44,8 @@ var loadBrunch = function(libPath) {
   // This approach is only needed for watch command.
   if ((cmd === 'w' || cmd === 'watch') && fs.existsSync(runCli)) {
     try {
-      return runSeparateWatchProcess(runCli);
+      runSeparateWatchProcess(runCli);
+      return;
```
denysdovhan: What's the difference? `runSeparateWatchProcess()` doesn't return anything. `loadBrunch` doesn't return anything too.

bin/brunch
```diff
@@ -51,7 +51,7 @@ var loadBrunch = function(libPath) {
     }
   }
 
-  // This is needed to support cases when the local brunch is of an older version
+  // This is needed to support cases when the local Brunch is of an older version
```
denysdovhan: 😱

lib/utils/plugins.js
```diff
@@ -130,7 +121,26 @@ const plugins = (config, craDeps) => {
     if (!hmrCompiler.supportsHMR) throw new BrunchError('HMR_PLUGIN_UNSUPPORTED');
   }
 
-  return getIncludes(plugins).then(includes => {
+  /* Get paths to files that plugins include. E.g. handlebars-brunch includes
+   * `../vendor/handlebars-runtime.js` with path relative to plugin.
+   */
+  const getIncludes = () => {
```
denysdovhan: Maybe it would be better to take `plugins` as an argument and not to use it from global scope.

bin/brunch
shvaikalesh: No

lib/utils/plugins.js
shvaikalesh: No

bin/brunch
shvaikalesh: Exactly. Having `return` there if it does not return anything is confusing.

PR #1663, opened 2017-02-03
Threaded workers

paulmillr:


paulmillr:
There's a bottleneck somewhere. Enabling PROCESS_FILE or increasing # of workers with/without enabling PROCESS_FILE slows it down.

Issue #1642, opened 2017-01-22
Killer feature request: Make Brunch STABLE


lydell:
## Description

The Brunch concept is super nice. Easy to install, easy to configure. Simple.

- Easy to use compilers.
- Great watch mode.
- Automatic npm installation.
- Awesome browser auto-reload.

Nothing super critical is missing from Brunch.

The issue is that Brunch keeps breaking. This is due to:

- Lack of tests.
- Lack of plugin API docs → plugins doing the wrong thing.
- Lack of npm releases and changelogs.

All of the time that I save from Brunch's super simple configuration is spent on hunting down the exact commits of Brunch and its plugins that work together and manages to compile my projects.

## Examples

- #1634: Brunch relying on private parts of the source-map package. This caused source maps to break when that package was updated, but there were non tests to catch this.
- #1621: A non-obvious feature that broke after a refactor of Brunch. A test could have caught it.
- #1546: Plugins breaking because of a mistake in a patch release. A basic integration test could have caught it.
- brunch/sass-brunch#120: Plugin being unsure of what it is supposed to return, resulting in no error messages from it. Better API documentation and some kind of "type-checking" tests could help.
- babel/babel-brunch#44: Same as the previous point.
- brunch/uglify-js-brunch#39: Plugin being unsure of what parameters it receives, and what it is supposed to return. Caused source maps to be broken.
- brunch/uglify-js-brunch@b0673d192da90dd4e4afa1eb2334935bad13a19c: Cleanup after plugin being uncertain of what it receives.

### Monorepo

Gather Brunch, deppack and all official Brunch plugins into a [monorepo](https://github.com/babel/babel/blob/master/doc/design/monorepo.md).

- Makes it easier to run tests that check if a change in Brunch breaks a plugin.

- Ability to share test setup between plugins. Currently, changes seem to be
  made in one plugin, and then things are copied to others. If a mistake is made in the first plugin's setup, that propagates. Making changes to all plugins is super tedious.

[...]

Issue #826, opened 2014-04-23
Improvement: control to modify how plugins are loaded


iamvdo:
The `on` and `off` options are really great, but i think of a small improvement when it comes to this config:

``` coffee
plugins:
    off: ['imageoptmizer-brunch']

overrides:
    production:
        plugins:
            on: ['imageoptmizer-brunch']
```

It's confused that this doesn't work. This syntax seems unusual, but it works:

``` coffee
plugins:
    off: ['imageoptmizer-brunch']

overrides:
    production:
        plugins:
            off: []
```

I don't know if it's the correct way of doing that, but that's the first one that came to me.

PR #1664, opened 2017-02-04
New architecture

paulmillr:


paulmillr:
It would be great to start introducing tests with the release. Want to make sure 2.11 is more stable than 2.10.

jacksonrayhamilton:
Looks like this needs rebasing, and tests.  Anything else to move this towards merging?

Also, if you could specify what needs tested, I could write some of these tests.  e.g., provide an outline of the specs you want, like:

- the compiler
  - should compile from one format to another
  - should compile from format A to B and then from format B to C

And so on.  Also, an overview of what this release accomplishes would be helpful.

shvaikalesh:
@jacksonrayhamilton This PR is not done yet. According to my time estimations, the fastest approach is to release breaking `3.0`. What is left is to replace `deppack` with `rollup` and make current tests pass, and then figure out how to make e2e tests a joy to write.

jacksonrayhamilton:
Thanks for the update @shvaikalesh.  Do you have an idea of what scenarios e2e tests might cover?

shvaikalesh:
@jacksonrayhamilton I will start with covering most error-prone (that was broken on my watch) logic, like merging `plugins.on|off` in `overrides` (see #826), merging `joinTo` and `entryPoints`, choosing plugins based on `env` etc. Hopefully, with some breaking changes in `3.0` there will be less such pain points.

lydell:
@jacksonrayhamilton There are also some ideas for tests in #1642.

rudi-c:
Hi @paulmillr @shvaikalesh, since this PR looks like a big one but currently isn't seeing progress, would it be possible to quickly write a summary of the features it will bring, API changes, and bugs it will fix? Right now Typescript is quite hard to use with Brunch and I have ideas on how to fix that, but I want to make sure to avoid doing duplicate and/or conflicting changes with the intended direction of Brunch.

explicitcall:
Hi @paulmillr @shvaikalesh, big +1 to @rudi-c comment, would be awesome to have some kind of roadmap for 3.0 release to understand what's coming. This would greatly help Brunch users in navigating existing issues and give better understanding if it's worth submitting new pull requests. 
Thanks in advance.

Tim15:
I am wondering if this will bring support for alternate bundlers like Rollup.

If we were to support Rollup, it would need some changes in the pipeline architecture. Since Rollup really shines with ES6 modules and tree-shaking, we would have some plugins run beforehand, and some (like Babel) run after the bundle is generated. Also, I wonder if Rollup would have better or worse performance than deppack. I'm not sure what types of optimizations they have done as far as atomic rebuilding.

It would be interesting to define a bundler API, that any new bundler could implement.

If I can help in any way, just ping me. I'm very excited to see what 3.0 brings!

StarpTech:
Hi, what's the status of this PR? The last commit is 1 year old and it seems inactive to me.

Tim15:
Any updates on this PR and version 3.0? Is the project still under development/being maintained?

jacksonrayhamilton:
@Tim15 I guess the entire Brunch team must have been eaten by wolves on the same night.  What a shame, I liked this project.

paulmillr:
Unfortunately we are focusing on more value creation; through other projects — at this particular moment. Sad we cannot spend more time with Brunch. Any help would be greatly appreciated.

kavdev:
@paulmillr has that changed at all? What needs to be done to get this merged?

clarknelson:
when are we getting this bad boy

paulmillr:
The status is simple: someone needs to take it over and finish the work (at least ~1 month 2h every day).

seanstrom:
@paulmillr What would be needed to work on to finish the work? Was there a feature list for 3.0, or a breaking change list?

paulmillr:
@seanstrom do you want to work on it? We can discuss this together. The idea is: an ultra-modern build tool that keeps Brunch simplicity, and increases it — while having everything required for 2020 web app dev.

seanstrom:
@paulmillr Yeah I would like to work on this. What's currently missing from Brunch that would be needed for 2020 app development? An example feature could be Code Splitting, or would that be something that is configured with a plugin?

paulmillr:
Code splitting indeed! We need to have first-class support for that. 

I think there should be some ways to simplify the config even more, and incorporate threading aka workers. This may increase the build speed 3x. We’d also want some kind of persistent file system caching, to increase speed of cold builds. 

One commonly requested feature is an ability to easily replace script and css paths inside index.html with git commit hashes or similar suffixes.

Cache & workers are simple. Conceptual changes, script path updates are harder.

paulmillr:
I want to release a version that keeps compatibility, while reducing the package size twice or more. See #1848. It will be named 3.0; the current PR is deferring until we figure out the requirements.

paulmillr:
Alright, we have a stable build in pull request #1848. Let's move over there.

PR #1756, opened 2017-10-10
Possible sourcemaps fix

shreeve:
This seems to work for me, but I'm not sure of the ramifications.

Anyone having problems with sourcemaps in Brunch 2.10, please check this out.

### DIFF
--- a/lib/fs_utils/generate.js
+++ b/lib/fs_utils/generate.js
@@ -86,7 +86,7 @@ const concat = (files, path, definitionFn, autoRequire, config) => {
 
   const processor = file => {
     root.add(file.node);
-    const data = file.node.isIdentity ? file.data : file.source;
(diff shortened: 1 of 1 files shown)


aMarCruz:
Works for me!
Thanks @shreeve .

shreeve:
FWIW, I don't think this is a robust solution, but it at least fixes the problem that I was seeing and apparently also @aMarCruz. It probably would be good to take a deeper dive and find a robust solution, but at least we should be getting closer... :-)

friday:
I think the real problem is that `_updateCache()` passes `file.sourceMap` instead of `file.map` to `_updateMap()`. As a symptom `file.node.isIdentity` is always true. See #1757

aMarCruz:
for me, #1757 solves minor issue with v2.10.11 (incorrect line offset) :+1:

PR #1776, opened 2017-12-22
Bump npm package depdendencies

PeterDaveHello:
Didn't update `ava` and `eslint` to the latest in this PR as they need more tuning.


paulmillr:
unsure if that won't break all the cases.

PeterDaveHello:
I think it depends on the coverage and the confidence of the tests :)

paulmillr:
#1849

PR #1777, opened 2017-12-22
Re-compress logo imgage losslessly

PeterDaveHello:
Use Google zopflipng to optimize the png image size, it's now less than 90% or origin one :)


paulmillr:
Thank you!

PR #1835, opened 2018-12-24
ref(entrypoints): Changed entrypoint error message

JosephShering:
Changed entrypoints warning to an error and made the error clearer
for the developer. Made it clearer how to fix the issue if
it is a problem


paulmillr:
thanks bro

PR #1848, opened 2019-08-26
v3.0

paulmillr:
Goals of this lightweight release:

1. Update dependencies to reduce the package size.
2. Create / Release new dep versions of anysort, micro-check-dependencies.
3. Remove features that are not common as per 2020 (bower, AMD etc.)
4. Prettify syntax by using async-await.
5. Bump node.js requirement to 10.8+.

Keeping everything else as-is.


paulmillr:
We should use https://github.com/zeit/micro for webserver I think

PR #1860, opened 2021-05-29
Readd plugins.order (fixes #1377)

### LINKED ISSUE #1377: Configure plugin order in config rather than rely on package.json order
It seems a bit fragile to have to rely on the order of declaration in package.json to determine the order that plugins are invoked in, given that npm --save reorders the dependencies alphabetically.  I'm basing this on brunch-guide, and a cursory glance over the source code.

It would be nice if this could be overridden in config.

On another note, I'd quite like to be able to add locally-defined plugins without having them be a package in node_modules.

reubano:
Reverts "remove plugins.order" (c759c745 from #1522)
Resolves #1377

### DIFF
--- a/lib/fs_utils/generate.js
+++ b/lib/fs_utils/generate.js
@@ -1,5 +1,6 @@
 'use strict';
 const debug = require('debug')('brunch:generate');
+const logger = require('loggy');
 const sysPath = require('universal-path');
 const anysort = require('anysort');
 const {promisify} = require('util');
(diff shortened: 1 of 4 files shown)


reubano:
[`plugins order`](https://github.com/brunch/brunch/pull/1860/files#diff-6855e3c14a839128c5aa991170a1f33776f2cbff6bdb5f18ce0bb41d80bde445R57) needs a test that actually checks the order.

paulmillr:
Thanks @reubano

Issue #1890, opened 2022-12-05
Directory structure incorrect in new project.


seanmorris:
### Description

The assets directory is not being handled correctly in a new project. 

### Expected behavior

The index.html file should be built to the public/ directory.

### Actual behavior

The index.html file is built to the public/assets/ directory.

### Environment

1. Brunch Version: 3.0.0 / 4.0.1
2. Node.js Version: 19.2.0
3. NPM Version: 8.19.3
4. Operating system: Debian GNU/Linux 11 Kernel 5.10.0-19-amd64

### `package.json` contents

```json
{
  "name": "brunch-app",
  "description": "Brunch.io application",
  "private": true,
  "author": "Brunch",
  "version": "0.0.1",
  "repository": "",
  "scripts": {
    "start": "brunch watch --server",
    "build": "brunch build --production"
  },
  "dependencies": {},
  "devDependencies": {
    "auto-reload-brunch": "^2",
    "brunch": "^4",
    "clean-css-brunch": "^3",
    "terser-brunch": "^4"
  }
}

```

### brunch config contents

none

seanmorris:
Working Lockfile: https://pastebin.com/UuipAhAK
Broken Lockfile: https://pastebin.com/z1iYS1ck

seanmorris:
It seems to be a bug with the new way strings are handled in anymatch 3.1.3. On line 50 of index.js, `false` is now provided as the second parameter to `normalizePath().`

https://github.com/micromatch/anymatch/blob/master/index.js#L50

Adding the following values to the root level `package.json` corrects the issue:

```json
{
  "overrides": {
    "brunch": {"anymatch": "3.1.2"},
    ".": {"anymatch": "3.1.2"}
  }
}
```

seanmorris:
https://github.com/micromatch/anymatch/issues/51

seanmorris:
I've found the issue, submitted a patch, and alerted the maintainers of the dependency to watch their semver.

https://github.com/micromatch/anymatch/issues/51

psagers:
It's worth noting here that a workaround was committed in a20d6a2997da1a192bed80f6d1f3fc2815196692 and released in 4.0.2 (Jan 10).

PR #1891, opened 2022-12-25
⚠️ Stripping trailing slash before matching conventions.

seanmorris:
https://github.com/brunch/brunch/issues/1890

It seems that anymatch 3.1.3 no longer automatically strips trailing slashes from paths before matching. This PR handles the stripping of the trailing slash before the match operation occurs.

Brunch won't work with new versions of anymatch without these changes, and the current package.json will not prevent new projects from pulling anymatch 3.1.3+. Thus nothing will work for new users until this is fixed.

**Also, without this patch, npm update will break existing projects.** This break may be silent as the index.html and other assets may already be built. They will NOT be correctly rebuilt without this patch. Furthermore, any CI/CD that depends on spinning this up from scratch will fail.

This patch seems to play nice with anymatch 3.1.1 and 3.1.2.

### DIFF
--- a/lib/fs_utils/asset.js
+++ b/lib/fs_utils/asset.js
@@ -11,10 +11,11 @@ const EMPTY_ARR = Object.freeze([]);
 
 // Returns all parent directories:
 // < parentDirs('app/assets/thing/index.html')
-// > ['app/', 'app/assets/', 'app/assets/thing/']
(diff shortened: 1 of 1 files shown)


paulmillr:
a20d6a2997da1a192bed80f6d1f3fc2815196692

